Issue #32, opened 2018-04-09
Guide to theming / custom themes


nolanlawson:
I should write a quick README about custom themes. Or maybe make it possible in the UI itself.

nolanlawson:
Partially fixed by https://github.com/nolanlawson/pinafore/pull/46

nolanlawson:
I'd say this is done basically; we have lots of contributed themes, so it seems it's easy enough to figure out

PR #46, opened 2018-04-09
docs: add guide to theming 

mlcdf:
It's a start.

(related to #32)


nolanlawson:
Awesome, thanks! And I see you noticed that the themes are a bit spread-out across the codebase. 😅 Something I need to fix, but in any case it looks like you got it exactly right.

PR #50, opened 2018-04-09
[WIP] Add dark Crowy theme

charlag:
Hi! Thanks for such an awesome thing!
This is my attempt on adding a dark theme.
I am completely open to feedback.
The only place where it sucks really badly is profiles and I'm not sure what to do about it, maybe there should be more variables.
screenshot from 2018-04-10 00-28-36
And name was chosen by random, I am okay with changing it!

### DIFF
--- a/routes/_static/themes.js
+++ b/routes/_static/themes.js
@@ -26,6 +26,10 @@ const themes = [
   {
     name: 'gecko',
     label: 'Gecko'
+  },
+  {
+    name: 'crowy',
+    label: 'Crowy'
   }
 ]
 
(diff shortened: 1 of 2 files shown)


charlag:
Oh, okay, there's also a problem with CW, it needs some work

nolanlawson:
Awesome start! I hadn't really gotten to dark themes yet, because I believe there may be some tweaks needed across various SCSS files to properly support them. We might also need to add some new CSS variables and modify some of the components to support them.

Also FWIW, there is now a guide to writing custom themes in case it helps: https://github.com/nolanlawson/pinafore/pull/46

nolanlawson:
FYI there is now a dark-style "Ozark" theme which we may be able to extend for this. :) https://github.com/nolanlawson/pinafore/pull/62

charlag:
Oh yes. I am not sure there's a demand for two dark themes and I don't have much time to work on this, maybe it's better to just close it...

PR #52, opened 2018-04-09
Fix #31: allow trailing slash in instance name

### LINKED ISSUE #31: Can't add an instance if there's a trailing slash in the URL
I tried to add `https://i.write.codethat.sucks/` as an instance and received this error:

`Error: NetworkError when attempting to fetch resource.. Is this a valid Mastodon instance?`

Nothing stands out in the dev console or network log (and I don't see a request attempt being made.)

Firefox ESR 52, Debian Linux 9. Is this maybe an issue with Service Workers? I see that entered Firefox in January 2018.

mlcdf:
TODO: write some tests.

(related to https://github.com/nolanlawson/pinafore/issues/31)


nolanlawson:
Awesome thank you, I think this is a pretty obvious fix so I'm happy to merge without a test. :)

PR #55, opened 2018-04-10
increase fetch timeout to 60 seconds for now

### LINKED ISSUE #29: 20 second timeout may be too aggressive
The goal is to fallback to offline mode when offline, but for some instances this means media uploads just don't work. Upping to 30 seconds or having a higher timeout for media would be a potential fix.

nolanlawson:
Fixes #29


nolanlawson:
In the future we should probably make read-only actions a shorter timeout than read-write actions (such as media uploads).

PR #62, opened 2018-04-10
Added 'Ozark' dark blue color theme

SpankyWorks:
This is based on a personal color scheme I use in code editors. Originally made for my own usage; I figured I would see if it was worth making a PR for.


nolanlawson:
Wow, this is awesome. Lots of folks are asking for a nice dark theme, and it looks like you've nailed it:

screenshot 2018-04-10 19 56 49

I think it might be useful to extract out some of these variables into a `_dark.scss` (for re-use with other dark themes), and it may also be neat to add dark scrollbars for WebKit/Blink browsers ala Mastodon, but this is such a good first stab that I'd be happy to add as-is. :) Thanks a lot for this!

PR #76, opened 2018-04-11
Increase in contrast for "--very-deemphasized" colors 

SpankyWorks:
Some have noticed that the very-deemphasized colors could use a little more contrast on dark themes, I just bumped up the lightness/opacity a little bit. They should still be darker than regular deemphasized colors.


nolanlawson:
Looks great to me. Thanks!

PR #78, opened 2018-04-11
add cobalt blue theme

chriswmartin:
Adds a new blue theme

screen shot 2018-04-11 at 10 40 55 am
screen shot 2018-04-11 at 10 41 49 am

### DIFF
--- a/routes/_static/themes.js
+++ b/routes/_static/themes.js
@@ -30,6 +30,10 @@ const themes = [
   {
     name: 'ozark',
     label: 'Ozark'
+  },
+  {
+    name: 'cobalt',
+    label: 'Cobalt'
   }
 ]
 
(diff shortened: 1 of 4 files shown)


nolanlawson:
(The tests appear to be failing for spurious reasons. I'll restart.)

This looks sweet; I love the yellow-on-blue. I feel like maybe the nav button text should be white instead of gray though? (Looks disabled/deemphasized.) I'll poke around with this later, but this feels like a really awesome start. :)

chriswmartin:
Good suggestion about the nav items, let me know what you think!

screen shot 2018-04-11 at 11 38 24 am

nolanlawson:
This is awesome! Reminds me of Mega Man. :)

My only minor nit is that I think the "Toot!" text and pencil icon shouldn't be quite so gray, so I lightened `$secondary-text-color` a bit. Other than that I love it, very chill. Thanks very much for this theme!

PR #79, opened 2018-04-11
allow scroll-to-top to be canceled by a user scroll

### LINKED ISSUE #72: manual scroll (or touchstart) should cancel scrolling to top
after clicking the current tab, trying to scroll back down while the ui is auto scrolling leads to fighting against the automatic scroll. it feels bad

this is even worse if the ui was already at the top, then it doesn't scroll visually but still prevents me from scrolling

codl:
this would fix #72

### DIFF
--- a/routes/_utils/smoothScrollToTop.js
+++ b/routes/_utils/smoothScrollToTop.js
@@ -13,7 +13,12 @@ const scroll = (node, key, target) => {
     const elapsed = Date.now() - startTime
     const percentage = elapsed / duration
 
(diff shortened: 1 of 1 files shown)


nolanlawson:
Awesome. I really like that you add the wheel/touchstart listeners at the beginning of the animation and then remove them at the end, to reduce their impact on scrolling. I'll give this a test in a bit and then merge. :)

nolanlawson:
Oh you also used passive event listeners. Excellent. :) Tested it out on mobile and desktop and looks great, thanks!

PR #84, opened 2018-04-11
Added a SCSS variable for the Account Profile header image overlay

SpankyWorks:
I tested it with all the themes currently in the project and it looks good with them all. Dark themes didn't play nicely with the transparent white.


SpankyWorks:
The variable I created in `_base.scss` does work for dark themes, at least for Ozark, because it currently includes the mixins for both `baseTheme()` and `darkTheme()`. If the future goal is to include only one of those mixins then it should also be updated in `_dark.scss`. Its default is set to a transparent version of the `$main-bg-color`, so it should be roughly as legible as toots are but with the cover image visible through it.

nolanlawson:
Ah okay, I missed that. Looks great! Only nit is that I feel the variable shouldn't be called `-webkit` because technically what I'm doing here is distinguishing browsers that support `backdrop-filter` (or its webkit variant) from those that don't. Today that means Safari and Edge vs Firefox and Chrome. (It has a nice frosted glass effect in Safari and Chrome. :))

Anyway I will make that tweak and then merge. This looks awesome.

SpankyWorks:
Ah, I see! I spend most of my time in FF so I saw the `webkit` in the style rule and immediately assumed it was a webkit-specific property.

nolanlawson:
Shipping in Edge soon and in Chrome behind a flag, but yeah I think originally it was Safari-only: https://caniuse.com/#feat=css-backdrop-filter

PR #108, opened 2018-04-14
add toot-cafe theme

mlcdf:
I also improve some contrasts compared to the original Toot Café theme.

screenshot-2018-4-14 pinafore home


nolanlawson:
This is awesome!! I tried to write "Majesty" to be TootCafé-esque, but this dark theme is way more accurate. 😁

To keep with the naming convention, though, is there a good name other than "Toot Café" you can think of? How about "Sorcery", since it kind of reminds me of the crystal ball emoji? :) 🔮

Issue #99, opened 2018-04-13
Focus on timeline on page load


jaranta:
This is a tiny usability issue, but I run into this every time I open Pinafore.

## Expected

After page load, Page Down, Page Up or Space buttons scroll the timeline.

## Actual

Page Down, Page Up or Space buttons don't scroll the timeline. They start working after mouse clicking the area around the timeline.

nolanlawson:
Good catch, looks like initially the focus is on the navigation instead of the timeline.

mlcdf:
I can work on this one.

mlcdf:
Not sure either. 

I guess the behaviour should be the same whether you are on the page displaying a timeline (local, federated, ) or not (settings, community, ...).

nolanlawson:
fixed in #116 and #147

PR #116, opened 2018-04-14
focus container on page load

mlcdf:
On page load, the focus is set the main content area instead of on the nav bar.

This fixes (#99).

### DIFF
--- a/routes/_components/Layout.html
+++ b/routes/_components/Layout.html
@@ -1,6 +1,6 @@
 <Nav :page />
 
-<div class="container">
+<div class="container" tabindex="0">
   <main>
     <slot></slot>
   </main>
@@ -17,7 +17,10 @@
 		components: {
 			Nav,
       InformationalFooter
-		},
+    },
+    oncreate () {
+      document.querySelector('.container').focus()
+    },
     store: () => store
 	}
 </script>
--- a/routes/_components/Layout.html
+++ b/routes/_components/Layout.html
@@ -1,6 +1,6 @@
 <Nav :page />
 
-<div class="container" tabindex="0">
+<div class="container" tabindex="0" ref:container>
   <main>
     <slot></slot>
   </main>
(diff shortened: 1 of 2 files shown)


nolanlawson:
Hmm, it looks like this is failing the tests because it will change the focus anytime `Layout.html` is created, which basically means anytime we navigate between different pages. In some cases we want the focus to be something else (e.g. you click a toot, you navigate from a timeline view to a thread view, you press the back button, and now you're still focused on the toot).

Is there a way to do this without JavaScript? If not, maybe the most practical thing is to put the `focus()` in `main.js` so that it only runs only when the very first page is loaded, and not during in-app navigations.

nolanlawson:
Another option: I just added a change to [`LazyPage.html`](https://github.com/nolanlawson/pinafore/blob/master/routes/_components/LazyPage.html) (which loads for all routes) to detect first load versus every subsequent load. We could use the same trick in `Layout.html`.

I may be best to do this in `Layout.html` because we don't even need to call `querySelectorAll('.container')`: you can use `ref:node` and then `this.refs.node` in Svelte to [get the node of a rendered element](https://svelte.technology/guide#refs).

mlcdf:
Thanks for the lead on this. Now, I'll try to make this work!

nolanlawson:
That's odd; the code looks correct to me but some tests are failing. I'll look into this when I get a chance!

mlcdf:
I tried to reproduce those failing tests locally, and it worked (it produced the expected result). So, I'm kinda lost on this one.

nolanlawson:
Hmmm, I can reproduce using Chrome on Mac with the command:

    npx testcafe --hostname localhost chrome --skip-js-errors tests/spec/010-focus.js

Oddly though I can't reproduce manually when using Chrome. It seems there may be some timing issue in a11y-dialog that causes it to think the previous focused element was `.container` rather than the `.play-video-button` when (clicking that button).

mlcdf:
Thank you for the support @nolanlawson!

PR #121, opened 2018-04-14
redirect from www.pinafore.social to pinafore.social

### LINKED ISSUE #61: www.pinafore.social should redirect to pinafore.social

nolanlawson:
Fixes #61


nolanlawson:
I'm worried this will break something, so holding off for a bit before deploying this.

PR #145, opened 2018-04-17
better JS size and async CSS with mini-css-extract-plugin

nolanlawson:
This uses [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin) to shave about 70kB off the total combined size of all JS chunks, by (gasp!) using actual CSS files instead of embedding everything in JS.

Unfortunately the older version of Sapper I forked doesn't support this, so I had to make [some modifications](https://github.com/nolanlawson/sapper/commit/da1d299662802263a4d608b332b3fc0f14f987e5). But I think it's well worth it for better perf.


nolanlawson:
[One extra modification](https://github.com/nolanlawson/sapper/commit/d9db963a055aaaf542d405619ca46eb120c6dd92) was required for my Sapper fork.

PR #147, opened 2018-04-17
Attempt to fix focus tests for #116

nolanlawson:
Attempt to fix #116 by explicitly focusing. If this works, we may want to expand this to all focus tests, because I suspect this may explain why the Firefox tests were failing as well.


nolanlawson:
Woot, passing! :)

nolanlawson:
rebase

PR #148, opened 2018-04-17
remove style-loader hack

nolanlawson:
I believe [this hack](https://github.com/nolanlawson/sapper/commit/c6da54184f734d16f61c967ae8121bf2ac824b36) is no longer needed since we're not using extract-text-plugin.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8405,7 +8405,7 @@
       }
     },
     "sapper": {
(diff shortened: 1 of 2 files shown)


nolanlawson:
checked prod version manually and this seems fixed

nolanlawson:
this removes the unnecessary style-loader from the bundle which was taking up a fair amount of space

PR #164, opened 2018-04-19
update and pin all deps

nolanlawson:
Since Pinafore is an app and not a library, it doesn't really make sense to have version ranges like ^4.7.4. The only thing that matters is what's in the lockfile, which is a locked version. So we've actually been running old versions of our deps for awhile.

OTOH I don't want stuff to randomly break, nor do I want our deps to slowly get out of date. So I'm just going to pin these and let Greenkeeper handle testing everything as updates roll in.


nolanlawson:
Incidentally there is already a test failure in the latest version of testcafe and node-sass. Not sure yet about testcafe, but node-sass is a bug on our end that needs to be fixed: https://github.com/sass/node-sass/issues/2336

PR #166, opened 2018-04-19
update testcafe to 0.19.2

nolanlawson:
I expect this to break the tests because it broke locally, but this is just to confirm


nolanlawson:
confirmed broken, should probably file a bug

nolanlawson:
I'm going to ignore this update for now, hopefully testcafe folks will fix it

PR #169, opened 2018-04-19
remove get() with string from Svelte calls

nolanlawson:
This has some odd failures that I'm currently working around (with `get() || {}`), but it's something we need to do because `get('foo')` is getting deprecated: https://github.com/sveltejs/svelte/pull/1347

### DIFF
--- a/routes/_actions/accounts.js
+++ b/routes/_actions/accounts.js
@@ -52,11 +52,10 @@ export async function clearProfileAndRelationship () {
 }
 
 export async function updateProfileAndRelationship (accountId) {
-  let instanceName = store.get('currentInstance')
-  let accessToken = store.get('accessToken')
+  let { currentInstance, accessToken } = store.get()
 
   await Promise.all([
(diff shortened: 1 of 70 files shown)


nolanlawson:
I don't know why `get()` sometimes returns null/undefined, but it seems to happen within `scheduleIdleTask()` which means it's happening outside of the component lifecycle, so that's probably it. I don't want to wrap every `get()` call ever though.

nolanlawson:
minimal svelte template to demonstrate what's causing the null bugs:

```html
<h1>Hello {{name}}!</h1>

<script>
	export default {
		oncreate() {
			setTimeout(() => this.destroy(), 1000)
			setTimeout(() => {
				let { name } = this.get()
				console.log('name', name)
			}, 2000)
		}
	}
</script>
```

nolanlawson:
bug filed on svelte: https://github.com/sveltejs/svelte/issues/1354

PR #190, opened 2018-04-20
simplify Status.html using spread params

nolanlawson:
This reduces the bundle size of Status.html from 31.63kB to 29.03kB while also making the code simpler and easier to read.


nolanlawson:
29.22 after the fix, but still an improvement

PR #193, opened 2018-04-20
use chunks:all in webpack splitChunks

nolanlawson:
This seems to result in better webpack chunks but needs verification


nolanlawson:
I played around with a few different constants, and I feel like these values offer the best tradeoffs, and the total bundle size is especially good (~606kB). Removing LodashWebpackPlugin `caching` saved about 3kB and is probably not necessary.

PR #204, opened 2018-04-21
make timeline components async

nolanlawson:
this should cut off 10-15kB from loading the home feed


nolanlawson:
321kB -> 311kB, I'll take it.

PR #214, opened 2018-04-22
[WIP] add lists of follows and followers

codl:
clicking on follow/follower counts shows list of follows or followers

i've tagged this wip because it's not super useful without user list pagination. i'm working on that too

### DIFF
--- a/routes/_components/profile/AccountProfileDetails.html
+++ b/routes/_components/profile/AccountProfileDetails.html
@@ -8,20 +8,24 @@
     </span>
   </div>
   <div class="account-profile-details-item">
-    <span class="account-profile-details-item-title">
-      Follows
-    </span>
-    <span class="account-profile-details-item-datum">
-      {{numFollowingDisplay}}
-    </span>
+    <a href='/accounts/{{account.id}}/follows'>
+      <span class="account-profile-details-item-title">
+        Follows
+      </span>
+      <span class="account-profile-details-item-datum">
+        {{numFollowingDisplay}}
+      </span>
+    </a>
   </div>
   <div class="account-profile-details-item">
-    <span class="account-profile-details-item-title">
-      Followers
-    </span>
-    <span class="account-profile-details-item-datum">
-      {{numFollowersDisplay}}
-    </span>
+    <a href='/accounts/{{account.id}}/followers'>
+      <span class="account-profile-details-item-title">
+        Followers
+      </span>
+      <span class="account-profile-details-item-datum">
+        {{numFollowersDisplay}}
+      </span>
+    </a>
   </div>
   <!-- TODO: re-enable this when we support profile editing -->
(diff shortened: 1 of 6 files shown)


nolanlawson:
Wow, thanks for jumping in on this. If possible, though, I'd love if we could tackle pagination separately, if for no other reason than to make it easier to review the PRs. 

Also I imagine pagination might be a bit tricky... once you get beyond 80 items, you _really_ ought to use a virtual list to reduce the DOM node count, but I'm not sure if either `VirtualList.html` or `PseudoVirtualList.html` are well-suited to this task. (Probably the pseudo one if anything, since it's much simpler.)

routes/_components/profile/AccountProfileDetails.html
```diff
@@ -8,20 +8,24 @@
     </span>
   </div>
   <div class="account-profile-details-item">
-    <span class="account-profile-details-item-title">
-      Follows
-    </span>
-    <span class="account-profile-details-item-datum">
-      {{numFollowingDisplay}}
-    </span>
```
nolanlawson: we might want to put the class on the `<a>` instead of the `<span>`, and probably remove the `<span>` altogether since it'll no longer be necessary

nolanlawson:
The Travis tests appear to have failed due to flakiness, so I restarted them

routes/_components/profile/AccountProfileDetails.html
codl: hmm there are two `span`s and two classes though? one for the title and one for the value. or did you mean the wrapping `div` and its class?

codl:
> If possible, though, I'd love if we could tackle pagination separately, if for no other reason than to make it easier to review the PRs.

Sure. I was planning to make a different PR

> once you get beyond 80 items, you *really* ought to use a virtual list to reduce the DOM node count

*Ah.* i don't really know how to work those 😅

nolanlawson:
Adding `PseudoVirtualList.html` shouldn't be _too_ tricky (see `Timeline.html` for usage), but I'd be happy to take that work on. :) In any case these changes look good; I would like to see tests and also fix the nit about `<a>` with unnecessary `<span>`, but I'm happy to fix that myself because I'd like to get this change in.

I'll merge and then fixup. Thanks for your hard work!

routes/_components/profile/AccountProfileDetails.html
nolanlawson: d'oh, my bad. you're right.

nolanlawson:
Merged in https://github.com/nolanlawson/pinafore/pull/225, thanks again! We can do the virtualization and the pagination in two separate PRs as needed. And maybe virtualization is overkill since the number of DOM nodes is so small anyway.

codl:
alright, sounds good. thanks!

PR #224, opened 2018-04-26
add service worker fixes to sapper

### LINKED ISSUE #80: "App update available" shows too frequently
In Firefox at least this seems to show even when there isn't actually an update. Probably I'm not detecting changes in the ServiceWorker lifecycle correctly.

nolanlawson:
This contains some tweaks to the service worker in http://github.com/nolanlawson/sapper which may or may not fix #80 but it's worth a shot.


nolanlawson:
Bah, just realized this won't work because I'm parsing the timestamp in some places to figure out which assets I can expire. This will be tricky to fix...

nolanlawson:
I've confirmed via manual testing that `dev.pinafore.social/service-worker.js` now returns `cache-control:max-age=0` and that the timestamp is always the same now regardless of which scaled node is serving it.

PR #226, opened 2018-04-27
use latest npm, update package-lock.json

nolanlawson:


nolanlawson:
On second thought, I'm nervous because this changes the package-lock.json file so much, and I'm not sure I can tell `now` to run the right npm version in production

PR #231, opened 2018-04-28
use latest npm@5 in CI

nolanlawson:


nolanlawson:
this doesn't actually seem to work in travis, so seems not worth the effort

PR #240, opened 2018-04-30
remove all instances of string coercion

nolanlawson:
Magical coercion from strings to ints will be gone in the next version of Svelte, so we need to be proactive about removing it. This was done with a regex, so hopefully nothing's broken...


nolanlawson:
probably better to get this in early rather than later

PR #244, opened 2018-04-30
slightly improve Svelte 2.0 compatibility

nolanlawson:
It seems that compatibility with Svelte 2.0 is going to be a multi-step process. We're too far away to do it all at once.

This gets us a little bit closer.


nolanlawson:
Most of this PR is about enabling `dev` warnings (which I realized were disabled) and then correcting all the warnings that Svelte is giving us, some of which will factor in to 2.0 compat.

PR #255, opened 2018-05-02
pass props to children in Status.html using `state` object

nolanlawson:
As of Svelte 2.0 we can use nicer code like this


nolanlawson:
not sure why it's failing, not worth pursuing

PR #275, opened 2018-05-07
use single transform:translate in virtual list

nolanlawson:
This fixes an issue with the autosuggestion overflow sometimes being rendered "under" the status underneath it, and it also makes the DOM do less work since we only have to update the `transform` on one DOM node instead of multiple.


nolanlawson:
this seems to cause some jerky rendering in Firefox so is maybe too ambitious

PR #294, opened 2018-05-13
try using ubuntu precise in travis

nolanlawson:
Chrome headless supposedly requires Ubuntu Trusty, but Trusty doesn't have ffmpeg which means we need to install a specific PPA package which adds 70 seconds to the build time. Let's see if we can avoid it.


nolanlawson:
my bad, I thought precise was newer than trusty

PR #313, opened 2018-05-22
Added Dockerfile

koyuawsmbrtn:
Note: A running Docker container must be killed instead of getting stopped. Ctrl-C/`SIGINT` doesn't seem to work while running Pinafore.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+# Using Alpine to keep the images smaller
+FROM alpine:latest
+
+# Pushing all files into image
+WORKDIR /app
+ADD . /app
+
+# Install updates and NodeJS+Dependencies
+RUN apk update && apk upgrade
+RUN apk add nodejs git python build-base clang
+
+# Upgrading NPM
+RUN npm i npm@latest -g
+
+# Install Pinafore and setting run-command
+RUN npm install
+RUN npm run build
+
+EXPOSE 4002
+
+CMD PORT=4002 npm start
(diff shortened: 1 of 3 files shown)


nolanlawson:
Thanks, this seems reasonable to merge now. Is the `SIGINT` issue a big deal? I don't know anything about Docker.

koyuawsmbrtn:
@nolanlawson admins just need to say `docker kill` instead of `docker stop`. This is not a big deal for now, yes.

nolanlawson:
Should this be documented? Is there some way to fix it in `server.js`?

koyuawsmbrtn:
It should be documented. Is there perhaps a handler, that fetches all keyboard interrupts or so?

nolanlawson:
It's possible to add SIGINT interceptors in node. Here is [one example of how it might be done](https://github.com/pouchdb/pouchdb-server/blob/fdc6ba740190b5bc178a2f6665e91910e70df459/packages/node_modules/pouchdb-server/lib/index.js#L304-L306). What exactly is the desired behavior here? That SIGINT kills the process?

koyuawsmbrtn:
This might do it. I'll try that.

nolanlawson:
Awesome, thank you. The error seems to be an intermittent failure, so I'll restart until it passes and then merge.

nolanlawson:
I don't understand why it's not passing. I'll keep looking into this.

nolanlawson:
Thank you again!

PR #318, opened 2018-05-24
attempt to fix flaky tests

nolanlawson:


nolanlawson:
Well, it passed the first time without even needing `travis_retry`, so that's a good sign.

PR #321, opened 2018-05-25
chore(package): update testcafe to version 0.20.2

nolanlawson:
Closes #300


nolanlawson:
testcafe still failing even with the roles removed :confused:

PR #339, opened 2018-05-26
travis already has postgres 9.1 by default

nolanlawson:


nolanlawson:
not working, apparently it doesn't have postgres by default, bizarre

PR #341, opened 2018-05-26
simplify travis postgres setup for mastodon

nolanlawson:
Supposedly Travis already has Postgres 9.1 installed by default which should work


nolanlawson:
this is not actually simpler and also not working

nolanlawson:
actually yeah let's keep trying this

nolanlawson:
Not working, not worth the effort to debug

PR #362, opened 2018-06-01
use npm install instead of npm ci in travis

nolanlawson:
I'm tired of builds constantly failing over this


nolanlawson:
actually if greenkeeper is not updating the lockfile then that's the fault of greenkeeper. fast builds are good builds

nolanlawson:
changed my mind, this is too annoying to deal with manually

PR #367, opened 2018-06-03
Greenkeeper: update now and esm

nolanlawson:


nolanlawson:
I'm also going to switch to caret-versioning instead of pinned versions in package.json. My reasoning:

- package-lock.json gets uploaded to now anyway, meaning we are effectively locked until package-lock.json is updated
- I cannot handle the frequency of updates from greenkeeper for every minor version bump. I really can't

PR #372, opened 2018-06-08
make tests faster by logging in directly

nolanlawson:
We can avoid having to actually touch the Mastodon web UI while logging in, at least for the majority of the tests.


nolanlawson:
This PR turned into "Nolan trying to make the tests faster and less flaky and throwing the kitchen sink at it"

PR #374, opened 2018-06-08
only install deps when running mastodon in travis

nolanlawson:


nolanlawson:
this seems too hard to get right due to postgres. but maybe I can just isolate the postgres stuff

nolanlawson:
doesn't seem to actually improve the time spent

PR #424, opened 2018-08-16
use page lifecycle

nolanlawson:


nolanlawson:
the point of this PR is to avoid using the `unload` event for non-essential state saves, as described in the excellent https://developers.google.com/web/updates/2018/07/page-lifecycle-api

PR #441, opened 2018-08-19
disable manifest.json for iOS

### LINKED ISSUE #45: PWA support on iOS 11.3
Reported [here](https://mastodon.social/@tbaldauf/99830914600338882):

> PWA functionality does not play nicely yet on iOS11.3 because when adding an instance, Safari opens & gets the login cookie instead of the sandboxed PWA webview, so it cannot login.

nolanlawson:
fixes #45

This PR UA sniffs for iOS devices and sends a 404 for `/manifest.json`. It's an unfortunate workaround for this problem, but I've confirmed that it allows "add to homescreen" to work on iOS. When you tap the homescreen icon, it just opens Pinafore in Safari.

### DIFF
--- a/server.js
+++ b/server.js
@@ -48,6 +48,17 @@ app.use(nonDebugOnly(helmet({
   }
 })))
 
+// TODO: remove this hack when Safari works with cross-origin window.open()
+// in a PWA: https://github.com/nolanlawson/pinafore/issues/45
+app.get('/manifest.json', (req, res, next) => {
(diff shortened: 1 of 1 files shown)


nolanlawson:
Dang, it looks like this will only work if you clear the Safari data. Apparently the way Service Workers work on Safari is that it doesn't overwrite a 200 response with a 404 response, so the client just never sees the 404ing manifest.json. 😕

Clearing Safari data works, though. Add to homescreen causes the icon to just open in Safari. I guess this is better than nothing.

PR #453, opened 2018-08-20
Add class names for notification/status types

dariusk:
I wanted to add user styles that hide certain notifications, so having class names by type is a big help for that. This commit simply appends the notification type to a "status-" class or a "notification-" class, depending on if it's a Status or Notification component. As a result you get classes like "status-mention", "status-favourite", and "notification-follow".

This could also eventually help with adding a Mastodon-style "filter your notifications by type" feature.

### DIFF
--- a/routes/_components/status/Notification.html
+++ b/routes/_components/status/Notification.html
@@ -3,7 +3,7 @@
           {status} {notification} on:recalculateHeight
   />
 {:else}
-  <article class="notification-article"
+  <article class={"notification-article notification-" + notification.type}
            tabindex="0"
            aria-posinset={index}
            aria-setsize={length} >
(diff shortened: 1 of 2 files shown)


routes/_components/status/Notification.html
```diff
@@ -3,7 +3,7 @@
           {status} {notification} on:recalculateHeight
   />
 {:else}
-  <article class="notification-article"
+  <article class={"notification-article notification-"+notification.type}
```
nolanlawson: nit: I like spaces around `+`s, but NDB

nolanlawson:
Ah actually I just had a thought: did you confirm that you can actually tweak the styles such that the notifications/statuses are hidden? Pinafore uses a virtual list, so I think by hiding them you might just create empty white space.

dariusk:
Ohhhh I see, it is absolute positioning based. Never mind!

nolanlawson:
Yeah sorry about that. :( I have work to do to get filtering in place. If only Mastodon had server-side filters...

PR #458, opened 2018-08-21
implement custom virtual list repeater

nolanlawson:
This is the first step in getting a more performant virtual list. I would like to get to where we are recycling DOM nodes and Svelte components inside of the virtual list (ala https://valdrinkoshi.github.io/virtual-scroller/), but as a first step I am basically reimplementing the Svelte `#each` loop and making a modest adjustment to defer `component.destroy()` work which based on my measurements is fairly expensive.

My measurements indicate that removing a component is most expensive, followed by adding a component, followed by updating. Updating is so cheap it's almost free. So this validates my assumption that moving to a model where we re-use all components rather than destroying/recreating would be a big perf win.

### DIFF
--- a/routes/_components/LazyImage.html
+++ b/routes/_components/LazyImage.html
@@ -22,34 +22,46 @@
 </style>
 <script>
   import { mark, stop } from '../_utils/marks'
+  import { observe } from 'svelte-extras'
 
   export default {
     oncreate () {
-      mark('LazyImage oncreate()')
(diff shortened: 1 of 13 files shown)


nolanlawson:
OK on second thought, I may just have the whole recycler working. I'll be really surprised if this actually works and is bug-free.

nolanlawson:
https://pinafore-mhzcjlhyox.now.sh

PR #459, opened 2018-08-21
implement custom virtual list repeater

nolanlawson:
less ambitious version of #458. instead of fully recycling nodes/components, for now we are just delaying the `destroy()` method in a requestIdleCallback to move some non-critical work out of the critical path.


nolanlawson:
TODO: make sure dom nodes retain the right order inside the list, for the benefit of screen readers

nolanlawson:
Yeah the tabs are jumping around all over the place, and positive tab indexes are not a good idea a11y-wise. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

PR #467, opened 2018-08-22
second attempt at modal back button

nolanlawson:
second attempt, here are the bugs with this approach:

- keeps adding entries to the history for every dialog you click, assuming you use something other than the back button to close them
- causes a re-render of the whole timeline component when you press the back button

overall this is just really not working. I guess I need to use the router that is actually built into Sapper.


nolanlawson:
another failed attempt at #60. this is surprisingly difficult

PR #473, opened 2018-08-23
WIP: use img.decode()

nolanlawson:


nolanlawson:
In theory this should reduce image decoding the main thread, but in practice we'll see if it helps much

PR #480, opened 2018-08-24
Add user meta fields to profile

### LINKED ISSUE #454: Display profile metadata fields
Added in Mastodon 2.4.0 (tootsuite/mastodon#6645, tootsuite/mastodon#7288)

CC #6

SpankyWorks:
Fixes #454 . My CSS grid and/or selectors might be kinda messy, but I think I managed to get it all working and looking nice.

screen shot 2018-08-23 at 20 39 04
screen shot 2018-08-23 at 20 39 38

### DIFF
--- a/routes/_components/profile/AccountProfileMeta.html
+++ b/routes/_components/profile/AccountProfileMeta.html
@@ -1,0 +1,93 @@
+<div class="account-profile-meta">
+   {@html meta}
+</div>
+<style>
+  .account-profile-meta {
+    grid-area: meta;
+    padding: 10px 0;
+    font-size: 0.9em;
+    word-wrap: break-word;
+    overflow: hidden;
+    white-space: pre-wrap;
+  }
+
+  :global(.account-profile-meta p) {
+    margin: 0 0 20px;
+  }
+
+  :global(.account-profile-meta p:first-child) {
+    margin: 0 0 20px;
+  }
+
+  :global(.account-profile-meta p:last-child) {
+    margin: 0;
+  }
+
+  :global(.account-profile-meta-field) {
+    display: grid;
+    grid-template-areas: "name value";
+    grid-template-columns: min-content auto;
+    align-items: center;
+    border-top: 1px solid var(--main-border);
+    padding: 5px 0px;
+  }
+
+  :global(.account-profile-meta-field:last-child) {
+    border-bottom: 1px solid var(--main-border);
+  }
+
+  :global(.account-profile-meta-field-name) {
+    grid-area: name;
+    padding: 5px 20px 5px 30px;
+    border-right: 1px solid var(--main-border);
+    font-weight: bold;
+  }
+
+  :global(.account-profile-meta-field-value) {
+    grid-area: value;
+    padding: 5px 30px 5px 20px;
+  }
+
+  @media (max-width: 767px) {
+    .account-profile-meta {
+      padding: 5px 0;
+    }
+
+    :global(.account-profile-meta-field) {
+      display: grid;
+      grid-template-areas: "name"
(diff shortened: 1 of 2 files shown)


nolanlawson:
Thank you, this looks awesome!! Looks like I may have bitrotted you with a recent commit, but I'll fix the merge conflict and then I'll be really happy to get this in!

routes/_components/profile/AccountProfileMeta.html
```diff
@@ -0,0 +1,132 @@
+<div class="account-profile-meta">
+   {@html meta}
+</div>
+<style>
+  .account-profile-meta {
+    display: grid;
+    grid-template-areas: "meta-name-1 meta-value-1"
+                         "meta-name-2 meta-value-2"
+                         "meta-name-3 meta-value-3"
+                         "meta-name-4 meta-value-4";
```
nolanlawson: I agree this `meta-name-1`/`-2`/`-3`/`-4` is a little messy, but it's not a big deal

routes/_components/profile/AccountProfileMeta.html
```diff
@@ -0,0 +1,132 @@
+<div class="account-profile-meta">
+   {@html meta}
```
nolanlawson: I think to avoid security issues with potential XSS attacks (as well as other issues with parsing HTML strings), it'd be more ideal to use a Svelte `{#each}` loop here

routes/_components/profile/AccountProfileMeta.html
```diff
@@ -0,0 +1,132 @@
...
+        let metaHTML = '';
+        for(var i=0; i < account.fields.length; i++) {
+          metaHTML += `<div class="account-profile-meta-field-name meta-name-${i+1}">`;
+          metaHTML += `<p>${account.fields[i].name}</p>`;
+          metaHTML += `</div>`;
+          metaHTML += `<div class="account-profile-meta-field-value meta-value-${i+1}">`;
+          metaHTML += `<p>${account.fields[i].value}</p>`;
+          metaHTML += `</div>`;
+        }
+        return metaHTML;
```
nolanlawson: once we use an `#each` loop we should be able to get rid of this computation

nolanlawson:
OK, I added the `#each` loop! :) and as a bonus, I made it so custom emoji work in fields.

screenshot 2018-08-23 18 56 59

Thanks again for your help!! I need to check to make sure there's no XSS risk from the emojified fields and I may also tweak the CSS a bit as well, but overall this looks amazing. Thanks so much!! :tada:

nolanlawson:
Had to open a separate PR since for some reason I couldn't commit to this branch ^

nolanlawson:
Looks like Mastodon already sanitizes field values.

screenshot 2018-08-23 19 08 37


Also Pinafore has CSP protection on inline scripts so it's not really an issue, but good to be safe.

SpankyWorks:
My unfamiliarity with Svelte has shown here, but it was fun to (finally) learn CSS grid to do this! Guess I'm gonna have to read up on Svelte some to take advantage of it next time I contribute.

nolanlawson:
No problem, thanks again for the contribution! :) Just a few pointers:

- `:global()` is only necessary when the classname isn't visible to Svelte from the HTML
- `grid-gap`, `grid-row-gap`, and `grid-column-gap` can avoid the need to do stuff like `:last-child` selectors to get the right margins in place. (I think the `:after` trick for the divider is really slick though)
- `grid-template-areas` isn't always required, in this case it's okay without it

I also decided after playing around for a bit to keep the 2-column view on both mobile and desktop, but you can see changes I made over in #482. :)

PR #501, opened 2018-08-27
fix application name

nolanlawson:
it should not show "Web" by default but should be absent if `application` is null


nolanlawson:
related: https://github.com/tootsuite/mastodon/issues/8476

Issue #503, opened 2018-08-27
Better accessibility for the profile fields


MarcoZehe:
Right now, the different fields and their values for each field are just divs without any further semantics. Some screen readers lump these together into an unintelligible string.

Suggestion:

1. Give each div with selector "account-profile-meta-name" a role of "term" and a unique ID.
2. Give the corresponding div with selector "account-profile-meta-value" a role of "definition" and an aria-labelledby with the idref of the "term" from step 1.

An alternative would be to use "table" for the container "account-profile-meta" and "gridcell" for each of the "account-profile-meta-cell" divs, but we are missing the rows here that would be required to use with valid WAI-ARIA parent/child/grandchild relations, so I am suggesting the above instead to get this read better and give it more semantic meaning.

nolanlawson:
Thanks, this was a really dumb oversight on my part! Unfortunately it would be tricky to use a `<table>` because I'm using CSS grid for this (probably overkill... as I'm not using anything other than a two-column layout).

I like the `role=term` and `role=description` approach as it seems to exactly describe what I want, while leaving the layout flexible in the future if I do decide to mix up the grid a bit. I found this definition in the WAI ARIA authoring practices, I'll give it a spin! https://www.w3.org/TR/wai-aria-1.1/#term

nolanlawson:
temp demo version here: https://pinafore-wrntcozavh.now.sh

nolanlawson:
fixed in https://github.com/nolanlawson/pinafore/pull/505

PR #505, opened 2018-08-28
improve a11y of fields, status page, and more

nolanlawson:
This should fix the accessibility bug described in https://github.com/nolanlawson/pinafore/issues/503 as well as a few other things. I used VoiceOver to validate my assumptions, so I hope that these improvements genuinely improve the screenreader experience.

Changes:

1. add `role=term` and `role=definition` to fields
2. Ensure that every page has an `<h1>` that is non-empty even if the visual content is empty for aesthetic reasons
3. Wrap the page header `<h1>` and back button in a `<nav>` with a label to distinguish it from the main `<nav>`
4. Avoid screen reader announcing the left arrow in the back button
5. Add `<h1>` heading for pinned statuses
6. Add `<h2>` headings inside of the profile to mark name and following, description, fields, and stats
7. Make some of the timeline `<h1>`s less chatty (for instance, don't announce the instance name for every single one, only the ones like home, notifications, local, and federated)
8. When a status or notification is focused, this is indicated with larger font size for sighted users, but for users of ATs I've added the word "focused" to the article label to show that the status is focused

/cc @MarcoZehe who may be able to provide guidance on whether I've made good choices or not!

### DIFF
--- a/routes/_components/timeline/PinnedStatuses.html
+++ b/routes/_components/timeline/PinnedStatuses.html
@@ -1,4 +1,5 @@
-<div role="feed" aria-label="Pinned toots" class="pinned-statuses">
+<h1 class="sr-only">Pinned statuses</h1>
+<div role="feed" aria-label="Pinned statuses" class="pinned-statuses">
   {#each pinnedStatuses as status, index (status.id)}
     <Status {status}
             timelineType="pinned"
--- a/routes/_components/timeline/PinnedStatuses.html
+++ b/routes/_components/timeline/PinnedStatuses.html
@@ -1,14 +1,16 @@
-<h1 class="sr-only">Pinned statuses</h1>
-<div role="feed" aria-label="Pinned statuses" class="pinned-statuses">
-  {#each pinnedStatuses as status, index (status.id)}
-    <Status {status}
-            timelineType="pinned"
-            timelineValue={accountId}
-            {index}
-            length={pinnedStatuses.length}
-    />
-  {/each}
-</div>
+{#if pinnedStatuses.length }
+  <h1 class="sr-only">Pinned statuses</h1>
+  <div role="feed" aria-label="Pinned statuses" class="pinned-statuses">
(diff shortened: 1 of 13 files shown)


nolanlawson:
temp demo version here: https://pinafore-wrntcozavh.now.sh

nolanlawson:
In VoiceOver, I don't see any difference for the fields with and without term/definition. For Narrator on Windows, it seems to read the term once for the term and once again for the definition, along with the word "definition". I suppose this is still more semantic though and at least makes the relationship clear.

MarcoZehe:
VoiceOver and Safari might not support term and definition yet, it‘s new in ARIA 1.1.

Does Narrator read the actual value, too, or is the text overridden by the term text? If the latter, the aria-labelledby would make more harm than good and should be removed. If the former, we‘re good.

The temp demo gives me an error 404, btw.

nolanlawson:
Narrator reads out the value as well. So e.g. "key: value" is read as "key" for the first, and then "key definition value" for the second.

I accidentally took the link down; here's a fresh one: https://pinafore-pgxzsthyth.now.sh

The headings could use some work (I know I probably shouldn't use multiple h1's per page) but I think it's an improvement over the old design.

routes/_components/timeline/PinnedStatuses.html
```diff
@@ -1,4 +1,5 @@
-<div role="feed" aria-label="Pinned toots" class="pinned-statuses">
+<h1 class="sr-only">Pinned statuses</h1>
```
MarcoZehe: This heading even appears when there are no pinned statuses in a profile. It's probably worth hiding this along with the div that is the container when there are no such statuses. Otherwise, the display will be a h1 with Pinned Statuses, followed by an H1 with regular statuses intro.

nolanlawson: Good catch! Will fix.

PR #517, opened 2018-08-29
run indexeddb operations in a web worker

nolanlawson:
This is just an experiment, needs validation to make sure it's actually a perf improvement

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2243,13 +2243,11 @@
             },
             "balanced-match": {
               "version": "1.0.0",
-              "bundled": true,
-              "optional": true
+              "bundled": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "bundled": true,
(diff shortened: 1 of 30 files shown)


nolanlawson:
#521 should fix this

nolanlawson:
preliminary testing in chrome on nexus 5 indicates that this does move some significant UI thread work to the background. we can always toggle it on and off as well if it becomes problematic.

screenshot 2018-08-28 20 35 07

screenshot 2018-08-28 20 36 51

nolanlawson:
all right, let's at least give this a whirl even if we disable it

nolanlawson:
this causes a weird visual effect where statuses loaded at the bottom of the timeline can load in the wrong order, causing a weird-looking cascading effect. May have to roll this back unless I can fix the ordering.

PR #520, opened 2018-08-29
reduce timeout when restoring test data

nolanlawson:
I'm tired of this test taking so long to run, maybe in latest Mastodon 100 milliseconds is okay.


nolanlawson:
welp that was a bust

PR #530, opened 2018-09-02
Make image use max-width and max-height in dialog

uiri:
Every time I click on an image to see the full version, it gets resized down to 300x200.

This PR fixes that :)

### DIFF
--- a/routes/_components/dialog/components/ImageDialog.html
+++ b/routes/_components/dialog/components/ImageDialog.html
@@ -10,14 +10,10 @@
       ariaLabel="Animated GIF: {description || ''}"
       {poster}
       {src}
-      {width}
-      {height}
     />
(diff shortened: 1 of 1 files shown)


nolanlawson:
Thanks, although I think there was a reason I had to explicitly set the height/width. I'll play around with this and see how it works.

nolanlawson:
Closing old issues, please reopen as necessary. Thanks!

nolanlawson:
To be clear, I couldn't understand what this PR was tackling or any examples of images that didn't look right in the current CSS. It's possible there's something I missed with some particular kind of images in certain servers, though.

uiri:
Hi, just adding a note here that it looks like the bug that I was observing has been fixed.

PR #538, opened 2018-09-06
remove workerize, use worker-loader+promiser-worker

nolanlawson:
Originally I wrote this because I was trying to fix webpack-bundle-analyzer (https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/193), but the problem still exists even while using vanilla `worker-loader`, so I'm not sure if I'll keep this PR or not.


nolanlawson:
This code is arguably more complex than what it replaces, and it doesn't fix https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/150 because we're seeing the same problems with both `workerize-loader` and `worker-loader`

PR #540, opened 2018-09-06
First stab at upgrading Sapper to latest

### LINKED ISSUE #416: Sync with latest Sapper
Hey — notwithstanding the [issues you've raised that are still open](https://github.com/sveltejs/sapper/issues/created_by/nolanlawson), I wanted to see what changes would be required to get Pinafore using sveltejs/sapper so that you don't need to maintain a fork.

It looks like [most of the changes](https://github.com/nolanlawson/sapper/commits/for-pinafore-7) should be fairly easy to backport, and some of them may no longer be relevant (for example, routes are no longer preloaded automatically, you have to [explicitly call `prefetchRoutes(...)`](https://sapper.svelte.technology/guide#prefetchroutes-routes-)). If I get time later I'll make individual issues for them on top of the existing ones.

I mention this because we just put out Sapper 0.15 which improves the routing quite a bit — I remember one of your bits of feedback was that it was annoying to have to recreate the nav etc from scratch upon navigation; routes are now hierarchical and allow persistent [layout](https://sapper.svelte.technology/guide#layouts) at any level (root or nested).

nolanlawson:
This would fix #416.

TODO:

- [ ] `index.html` isn't generated anymore (https://github.com/sveltejs/sapper/issues/422)
- [ ] set custom `cache-control` headers _(seems we can just fork to achieve this)_
- [x] set `SAPPER_TEMPLATE` to avoid service worker bugs when deploying to now.sh (https://github.com/sveltejs/sapper/pull/423)
- [x] Sapper has inline scripts now, this is messing with CSP (https://github.com/sveltejs/sapper/pull/424)

Bugs:

- [ ] CSS styles seem messed up on the "add.html" page, spinner is the wrong color
- [ ] When logging out of an instance, fails to redirect back
- [ ] Warning in webpack style-loader about ambiguous order between CSS files (is this a big deal?)

Nice to have:

- [ ] disable service worker on localhost vs 127.0.0.1? (this was nice for debugging but maybe should be removed)
- [ ] default port should be 4002, but it should be user-overridable via the `PORT` env variable


nolanlawson:
this would fix #416

nolanlawson:
This is not going to be possible until https://github.com/sveltejs/sapper/issues/422 is fixed. I've looked into it and I don't think I have the chops to implement it. 😞

PR #541, opened 2018-09-06
Inherit colors for input and textarea

sorin-davidoi:
When using a dark GTK theme `input` and `textarea` use the theme's colors by default. This makes sure that they inherit the page's colors.

# Current behavior with the Arc GTK theme
screenshot from 2018-09-07 00-00-09

Inheriting the page's color might not be desirable (see following screenshot). If that is the case we can just set it to white background with black text.

# Behavior with this patch (Arc GTK theme, Ozark Pinafore theme)
image

### DIFF
--- a/scss/global.scss
+++ b/scss/global.scss
@@ -82,6 +82,12 @@ input {
   box-sizing: border-box;
 }
 
+// Fixes gray/black background when using a dark GTK theme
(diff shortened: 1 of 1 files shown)


nolanlawson:
LGTM, thanks!

PR #543, opened 2018-09-07
Make external links open in new tab

uiri:
External links inside toots open up in a new tab (using target='_blank') instead of the same tab.

### DIFF
--- a/routes/_components/status/StatusContent.html
+++ b/routes/_components/status/StatusContent.html
@@ -114,6 +114,9 @@
           } else if (anchor.getAttribute('rel') === 'nofollow noopener') {
             anchor.setAttribute('title', href)
           }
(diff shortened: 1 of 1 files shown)


zcdunn:
Thanks! This will make using Pinafore in Firefox's sidebar way better

nolanlawson:
In which cases do external links not already have `target="_blank"`? I just checked a few toots, and it seems that all the external ones already have `target=_"blank"` on them. Which version of Mastodon are you using, and can you link to a toot that demonstrates this?

nolanlawson:
Closing old issues, please reopen as necessary. Thanks!

uiri:
I'm using Pleroma. It seems that, unlike Mastodon, Pleroma does not automatically insert `target="_blank"` into links and leaves that to the client to implement.

Could you please re-open and reconsider this PR?

PR #548, opened 2018-09-22
update several deps

nolanlawson:


nolanlawson:
looks like this will have to be done piecemeal to find the breakage

PR #579, opened 2018-09-28
Push notifications

### LINKED ISSUE #19: Push notifications
I'm not sure we could do this in a privacy-respecting way. I'd really prefer not to handle any user data server-side.

Perhaps it would be possible to extend the Mastodon server API to allow clients to hook into the web push notification system? That way all user data would still reside with the instance, and the web client would just be responsible for receiving pushes.

sorin-davidoi:
Basic implementation, without ~actions,~ grouping or the complicated logic that Mastodon has for focusing existing clients when clicking the notification.

I'm not sure how this will work with multiple instances (especially when opening a new tab via the notification that is not from the current instance). I think that would need a way to pass the desired instance via the URL (as a query parameter?).

TestCafe now supports [HTTPS](https://devexpress.github.io/testcafe/documentation/using-testcafe/common-concepts/connect-to-the-testcafe-server-over-https.html), so if that is set up I could write some basic tests to make sure that settings are saved in the backend.

No feature-detection is currently done. How would you prefer this to be handled - hide the settings entirely, disable them? I personally would like to disable them, informing the user that their browser doesn't support the feature.

Closes #19.

### DIFF
--- a/routes/_pages/settings/instances/[instanceName].html
+++ b/routes/_pages/settings/instances/[instanceName].html
@@ -15,6 +15,25 @@
         <AccountDisplayName account={verifyCredentials} />
       </span>
     </div>
+    <h2>Push notifications:</h2>
+    <div class="push-notifications">
+      {#if $notificationPermission === "denied"}
+        <p role="alert">You have denied permission to show notifications.</p>
+      {/if}
+      <form id="push-notification-settings" disabled ref:pushNotificationsForm aria-label="Push notification settings">
+        <input type="checkbox" id="push-notifications-follow" name="follow" on:change="onPushSettingsChange(event)">
+        <label for="push-notifications-follow">New followers</label>
+        <br>
+        <input type="checkbox" id="push-notifications-favourite" name="favourite" on:change="onPushSettingsChange(event)">
+        <label for="push-notifications-favourite">Favourites</label>
+        <br>
+        <input type="checkbox" id="push-notifications-reblog" name="reblog" on:change="onPushSettingsChange(event)">
+        <label for="push-notifications-reblog">Boosts</label>
+        <br>
+        <input type="checkbox" id="push-notifications-mention" name="mention" on:change="onPushSettingsChange(event)">
+        <label for="push-notifications-mention">Mentions</label>
+      </form>
+    </div>
     <h2>Theme:</h2>
     <form class="theme-chooser" aria-label="Choose a theme">
         <div class="theme-groups">
@@ -73,6 +92,17 @@
   }
   .acct-display-name {
     grid-area: display-name;
+  }
+  .push-notifications {
+    background: var(--form-bg);
+    border: 1px solid var(--main-border);
+    border-radius: 4px;
+    display: block;
+    padding: 20px;
+    line-height: 2em;
+  }
+  .push-notifications p {
+    margin: 0;
   }
   .theme-chooser {
     background: var(--form-bg);
@@ -148,8 +178,10 @@
     logOutOfInstance,
     updateVerifyCredentialsForInstance
   } from '../../../_actions/instances'
+  import { updatePushSubscriptionForInstance, updateAlerts } from '../../../_actions/pushSubscription'
   import { themes } from '../../../_static/themes'
   import AccountDisplayName from '../../../_components/profile/AccountDisplayName.html'
+  import { toast } from '../../../_utils/toast'
 
   export default {
     async oncreate () {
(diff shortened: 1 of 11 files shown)


templates/service-worker.js
```diff
@@ -94,3 +94,123 @@ self.addEventListener('fetch', event => {
...
+
+async function showSimpleNotification(data) {
+  await self.registration.showNotification(data.title, {
+    icon: data.icon,
+    body: data.body
+  })
+}
+
+async function showRichNotification(data, notification) {
+  switch (notification.type) {
```
sorin-davidoi: These are very similar right now but I prefer to handle them separately since they may diverge in the future (it is also clearer this way).

templates/service-worker.js
```diff
@@ -94,3 +94,123 @@ self.addEventListener('fetch', event => {
...
+self.addEventListener('notificationclick', event => {
+  event.waitUntil((async () => {
+    if (event.notification.data && event.notification.data.url) {
+      await self.clients.openWindow(event.notification.data.url)
+      await event.notification.close()
+    }
+  })())
+})
+
+// Copy-paste from ajax.js
```
sorin-davidoi: Maybe it would make sense to process the file with Webpack so that imports work?

templates/service-worker.js
```diff
@@ -94,3 +94,123 @@ self.addEventListener('fetch', event => {
     return fetch(req)
   })())
 })
+
+self.addEventListener('push', event => {
+  event.waitUntil((async () => {
+    const data = event.data.json()
+    const { origin } = new URL(data.icon)
```
sorin-davidoi: Can't think of a better option to get the instance from the notification payload.

routes/_pages/settings/instances/[instanceName].html
```diff
@@ -159,6 +191,15 @@ <h2>Theme:</h2>
...
       await updateVerifyCredentialsForInstance(instanceName)
+      await updatePushSubscriptionForInstance(instanceName)
+
+      const form = this.refs.pushNotificationsForm
+      const { pushSubscription } = this.store.get()
+
+      form.elements.follow.checked = pushSubscription && pushSubscription.alerts && pushSubscription.alerts.follow
+      form.elements.favourite.checked = pushSubscription && pushSubscription.alerts && pushSubscription.alerts.favourite
+      form.elements.reblog.checked = pushSubscription && pushSubscription.alerts && pushSubscription.alerts.reblog
+      form.elements.mention.checked = pushSubscription && pushSubscription.alerts && pushSubscription.alerts.mention
```
Comment (nolanlawson) [inline routes/_pages/settings/instances/[instanceName].html]: There should be a way to bind this to the store data without needing to use `refs`. The checkboxes for themes use a similar system.

templates/service-worker.js
nolanlawson: This works fine AFAICT1

templates/service-worker.js
nolanlawson: :+1:

templates/service-worker.js
nolanlawson: SW will be handled by Webpack once I upgrade us to the latest version of Sapper, but unfortunately I'm blocked on that for now: https://github.com/nolanlawson/pinafore/issues/416

nolanlawson:
This looks amazing to me, thanks for the PR! Other than the form data-binding issue (i.e. using Svelte's built-in two-way binding rather than `refs`) and how to handle unsupported browsers, this looks great to me.

Testing with TestCafé would be nice, but it's not a requirement for this feature from my perspective. (Setting up HTTPS for tests seems cumbersome to me at hte moment.)

As for how to handle unsupported browsers, I agree that a message telling folks that their browser doesn't support push notifications would be the best approach.

One other thing I note: folks will have to re-authentication with their instances, correct? Because I forgot the `push` permission when registering before.

Thanks again! This is a much-anticipated feature. :)

routes/_pages/settings/instances/[instanceName].html
Comment (sorin-davidoi) [reply routes/_pages/settings/instances/[instanceName].html]: I tried to avoid using `bind:checked` since the changes are not applied instantly (like the themes), since we need to perform a network request that may fail (handling that would mean having to revert the state of the store, which implies keeping track of the previous state of `store.pushSubscription`).

sorin-davidoi:
> One other thing I note: folks will have to re-authentication with their instances, correct? Because I forgot the push permission when registering before.

I think we can detect this case from the response and show a toast informing the user that they need to re-authenticate.

routes/_pages/settings/instances/[instanceName].html
Comment (nolanlawson) [reply routes/_pages/settings/instances/[instanceName].html]: Hmm, in that case I would still prefer to handle it with state than with direct DOM manipulation. In the past for these kinds of "optimistic" UI updates, I have used state overrides. For instance, you can set a boolean value on the local component state that overrides the store and just updates the checked state. I'll make the modification so you can see what I mean :)

Comment (nolanlawson) [reply routes/_pages/settings/instances/[instanceName].html]: Hmm, I am playing around with this, and the code is pretty gnarly. :( Maybe it will be best to just merge your PR as-is, since everything else looks fantastic to me. (Frankly I am amazed you managed to figure out my idiomatic code as much as you did! I didn't document everything nearly as well as I could have.)

nolanlawson:
This looks great overall, thanks so much! Excited to get this out in the next release. :)

routes/_pages/settings/instances/[instanceName].html
Comment (sorin-davidoi) [reply routes/_pages/settings/instances/[instanceName].html]: Apart from my unfamiliarity with Svelte I didn't have major issues navigating the code, looks very well structured to me :)

PR #581, opened 2018-10-06
refactor instance settings

nolanlawson:
The "instance settings" screen was getting a little hairy, so I broke it up into multiple small components.

### DIFF
--- a/routes/_components/settings/instance/PushNotificationSettings.html
+++ b/routes/_components/settings/instance/PushNotificationSettings.html
@@ -1,0 +1,93 @@
+<div class="push-notifications">
+  {#if pushNotificationsSupport === false}
+  <p>Your browser doesn't support push notifications.</p>
+  {:elseif $notificationPermission === "denied"}
+  <p role="alert">You have denied permission to show notifications.</p>
+  {/if}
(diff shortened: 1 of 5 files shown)


routes/_components/settings/instance/PushNotificationSettings.html
```diff
@@ -0,0 +1,93 @@
...
+
+  export default {
+    async oncreate () {
+      let { instanceName } = this.get()
+      await updatePushSubscriptionForInstance(instanceName)
+
+      const form = this.refs.pushNotificationsForm
+      const { pushSubscription } = this.store.get()
+
+      form.elements.follow.checked = pushSubscription && pushSubscription.alerts && pushSubscription.alerts.follow
```
sorin-davidoi: Updating `.checked` after `updatePushSubscriptionForInstance` leaves the checkboxes uncheked for the duration of the network call (didn't noticed it during development :( ). Could we perhaps move this into a function and call it before and after `updatePushSubscriptionForInstance` until a better solution is found?

nolanlawson: Seems reasonable, yeah. I tried to think of a better solution, but couldn't for now. :(

PR #585, opened 2018-10-16
Provide directions to change the theme

phtan:
When I wanted to change the theme, the page 'Settings > General' was the first place I looked  
Not managing to find it, I went on a search throughout various issues and PRs (#46  , for
example) before locating it.

Now I hope to spare others the search

### DIFF
--- a/routes/_pages/settings/general.html
+++ b/routes/_pages/settings/general.html
@@ -24,7 +24,11 @@
       <label for="choice-omit-emoji-in-display-names">Remove emoji from user display names</label>
     </div>
   </form>
-
+  
+  <h2>Themes
+  </h2>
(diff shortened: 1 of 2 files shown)


routes/_pages/settings/general.html
```diff
@@ -24,7 +24,11 @@ <h2>UI Settings</h2>
       <label for="choice-omit-emoji-in-display-names">Remove emoji from user display names</label>
     </div>
   </form>
-
+  
+  <h2>Themes
+  </h2>
+  <p>Changes to the theme (colours) of Pinafore can be made in <i>Settings &gt; Instances &gt; *your instance*</i>
+  </p>
```
nolanlawson: I like this change, but I'm not so sure about the second one. Thank you very much for this PR!

phtan: As you see fit.

It's my pleasure.

PR #625, opened 2018-11-11
avoid delaying when inserting every post in tests

nolanlawson:
This should speed up the tests by avoiding artificial delays. But I'm going to do 10 runs to make sure it actually doesn't introduce flakiness.


nolanlawson:
10 runs, 10 greens. LGTM

PR #627, opened 2018-11-11
chore(travis): faster ffmpeg install in travis

nolanlawson:
Just download a binary and cache it instead of using an apt package.


nolanlawson:
seems impossible to download ffmpeg binaries in travis. too bad

PR #628, opened 2018-11-11
chore(travis): use yarn

nolanlawson:
Mastodon is already using yarn anyway. If yarn is faster than npm in Travis, then I'll use it.


nolanlawson:
[before](https://travis-ci.org/nolanlawson/pinafore/builds/453632859) and [after](https://travis-ci.org/nolanlawson/pinafore/jobs/453647301)

test suite 0: 14m10s -> 13m56s
test suite 1: 19m10s -> 16m59s

`npm ci`: 53.45s
`yarn`: 37.26s

OK yeah it seems faster, and this way we are caching less because we're using the same system as Mastodon. I need to update the docs to specify to use `yarn` rather than `npm` though, but this looks good.

nolanlawson:
On second thought, I wonder if the difference is just because I also updated the cache system. Worth testing.

nolanlawson:
I'm skeptical of the added complexity and breaking change, and also based on multiple runs, `npm ci` seems roughly as fast as `yarn`. Gonna go with #632 instead

PR #630, opened 2018-11-11
chore(travis): update to mastodon v2.6.1

nolanlawson:


nolanlawson:
Something definitely changed in 2.6.1 and tests seem legit broken.

nolanlawson:
I'm not happy about removing tests for direct messages, but since direct messages don't seem to appear in the home timeline anymore, it's difficult to test without creating a new DM interface (which is I guess what we should do).

PR #633, opened 2018-11-11
chore(travis): try explicit apt-get installs for faster scripts

nolanlawson:
I have a hunch this may be faster given 1) we have to manually install ffmpeg anyway, and 2) we don't want to run any of this stuff for the "deploy" job


nolanlawson:
I don't trust all this complexity I'm introducing

PR #637, opened 2018-11-12
chore(travis): try travis sudo:false infrastructure

nolanlawson:
Supposedly these boot faster and have IPv6, which should avoid the Redis issue.


nolanlawson:
Redis is still busted, and this will probably be slower overall anyway based on https://git.io/fpqRe

PR #650, opened 2018-11-16
perf(virtuallist): add contain:content to virtual list items

nolanlawson:
As far as I can tell, CSS containment really does improve performance of layout/style calculation in Chrome. I tested the "Favorites" page by loading it with the Chrome Dev Tools profiler running, then measured the total "Rendering" costs (the purple stuff, really Style/Layout). Based on the median of 5 runs, this reduces those costs from 142.2ms to 103.2ms. I'll take it!


nolanlawson:
Adding `contain:content` to the nav bar and the main container whittles it down even further, to 85.5. This is like free performance sauce.

Issue #219, opened 2018-04-24
Scrolling is all sorts of messed up on some browsers


Granitosaurus:
There seems to be some weird scrolling issues going on some browsers.

For example I'm running latest Qutebrowser and scrolling tends to stop working at random.   
To add to that scrolling doesn't work when loading the page for the first time - you need to select toot are for it to start working.

nolanlawson:
Qutebrowser is WebKit-based, right? Or do you have the update where it's Chromium-based? Also I assume this is Linux desktop, and you're scrolling with the mousewheel or keyboard or some other non-touch method?

There is a similar issue for Safari, which is also Webkit: #123. I wonder if it's related.

Scrolling ought to work when you first load the page since #99 was fixed; not sure why it's not in your case.

Granitosaurus:
Qutebrowser is webengine based.  

I cannot replicate or figure out what the issue it seems to stop completely randomly - sometimes it's fine for several pages. 

To add to that other scroll functions don't work either like scroll by percantage. Seems like qutebrowser is confused by header + content box structure pinafore is using :|

nolanlawson:
Since Qutebrowser is WebKit-based, this may be related to #123.

Granitosaurus:
Qutebrowser is based on webengine (chromium). There is support for webkit but webkit being deprecated makes it pretty much dead and useless.

nolanlawson:
I'm going to assume this is fixed by https://github.com/nolanlawson/pinafore/pull/656

Issue #652, opened 2018-11-16
Scrolling is janky on Chrome for Android


nolanlawson:
It's actually smoother in Firefox, and also smoother in the main Mastodon frontend in Chrome. It appears to be some kind of rendering/layers issue. Maybe also related to https://github.com/nolanlawson/pinafore/issues/526

nolanlawson:
fixed by https://github.com/nolanlawson/pinafore/pull/656

PR #656, opened 2018-11-17
fix(scrolling): use body as scrolling container

### LINKED ISSUE #526: Make body the scrolling container
Would enable mobile browsers to hide the address bar while scrolling down.

nolanlawson:
Fixes #526

I have not thoroughly tested in all browsers yet, but I will.


nolanlawson:
now link: https://pinafore-mkuavsjlob.now.sh

nolanlawson:
As a bonus, there is apparently no more reason to explicitly focus the container on first load. (#116)

nolanlawson:
This also seems to fix scrolling perf issues in Chrome. I guess because Chrome for Android hasn't optimized scrolling performance of subscrollers yet. (#652)

Maybe this will also fix #219

PR #661, opened 2018-11-19
fix(emojos): fix emojos on Ubuntu and Chrome on Windows

nolanlawson:
With this PR we wrap emojis in a `<span>` with a special class to try to mark them with system emoji fonts. This is the best method I could figure out, testing on various browsers and OS's:

- Firefox+Chrome+Epiphany on Ubuntu 18.04
- Chrome+Firefox on Android 6
- Chrome+Firefox on Android 8
- Safari on iOS
- Safari+Firefox+Chrome on macOS
- Edge+Chrome+Firefox on Windows 10

I definitely didn't catch all the cases, and some of the emoji still don't show up correctly, but I think this is close to the best I can do without using my own custom emoji (as Mastodon does).

Firefox on Ubuntu 18.04:

screenshot from 2018-11-18 20-17-14

Chrome on Ubuntu 18.04:

screenshot from 2018-11-18 20-16-21

Epiphany on Ubuntu 18.04:

screenshot from 2018-11-18 20-18-06


nolanlawson:
Epiphany browser is fixed now, and in case I didn't mention it, Chrome on Windows is also fixed now (for the "woman bowing" emoji, haven't tested all of them).

PR #678, opened 2018-11-22
test: make tests less flaky

nolanlawson:


nolanlawson:
The tests are still flaky, but this is better

PR #685, opened 2018-11-24
feat(design): add custom scrollbars for each theme

### LINKED ISSUE #683: Add custom scrollbars for each theme
This is supported in Chrome, Firefox, and Safari via either the `-webkit` proprietary CSS or the new `scrollbar-color` CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color

nolanlawson:
fixes #683


nolanlawson:
I'd say these look pretty good

out

nolanlawson:
That pink one could maybe use a bit more contrast, but it's the only one, and it'd be tough to do something special just for one theme.

nolanlawson:
nah, sass `!default` made it easy, went ahead and improved the contrast

PR #690, opened 2018-11-25
fix(a11y): improved aria-label for status and notifications

### LINKED ISSUE #689: Improve aria labels for statuses and notifications
For reference: https://github.com/tootsuite/mastodon/issues/8192

I see no problem with adding aria labels like what Mastodon FE currently has.

nolanlawson:
fixes #689 

I made a slight modification to the Mastodon frontend's system by adding "Public"/"Unlisted"/"Followers-only"/"Direct" to the end of the `aria-label`, since this seems like useful information and is only subtly indicated elsewhere. (E.g. in the fact that you can't boost a direct/followers-only status.)

This adds a bit of a perf hit because we now need to use `DOMParser` to parse HTML to strings, but I measured it, and it seems to be measurable but not huge (on the order of 0.5ms -> 1.5ms per toot on my desktop when running in Chrome at 4x slowdown). We also take a slight perf hit because we are now parsing dates twice, but this is also fairly negligible (0-1ms per date formatted). I'm not super concerned since #656 has moved scrolling off-main-thread for all browsers.

We still don't have `j`/`k` keyboard scrolling, so I admit I'm not sure how useful this is to screen reader users. I also don't have a good screen reader to test on right now, since my Mac is busted. But hopefully this helps.

### DIFF
--- a/tests/spec/118-display-name-custom-emoji.js
+++ b/tests/spec/118-display-name-custom-emoji.js
@@ -11,7 +11,7 @@ import {
   settingsNavButton,
   sleep
 } from '../utils'
-import { updateUserDisplayNameAs } from '../serverActions'
+import { postAs, updateUserDisplayNameAs } from '../serverActions'
 import { Selector as $ } from 'testcafe'
 
 fixture`118-display-name-custom-emoji.js`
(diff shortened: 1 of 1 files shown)


nolanlawson:
OK, we are now only formatting dates once

PR #696, opened 2018-11-26
fix(a11y): default "prefers reduced motion" to OS/browser default

### LINKED ISSUE #695: Default the "reduce UI animations" preference to true if prefers-reduced-motion is true
The default should be

```js
window.matchMedia('(prefers-reduced-motion: reduce)').matches
```

For users who have configured their OS/browser this way.

nolanlawson:
fixes #695


nolanlawson:
https://pinafore-zyvpksdyuy.now.sh

nolanlawson:
Tested in Safari and it correctly defaults, but you can still override it.

PR #702, opened 2018-12-01
fix(a11y): fix NVDA crash on long aria-label

### LINKED ISSUE #694: New aria-labels cause NVDA to crash
N.B.: https://github.com/nvaccess/nvda/issues/8980

Apparently the change in #689 causes NVDA to crash. This needs to be investigated before deploying to prod.

nolanlawson:
fixes #694


nolanlawson:
https://pinafore-efqmlfsgov.now.sh

nolanlawson:
This does indeed seem to fix it, unless I missed a super-long toot that would still cause a crash.

Issue #673, opened 2018-11-21
Sticky-positioned button not sticking in iOS Safari


nolanlawson:
This is hopefully the last regression introduced by the "document as scroller" fix. Probably has something to do with overflow-x:hidden

nolanlawson:
Hm, so it turns out that this is caused by putting `overflow-x:hidden` on the `.main-content` element. Putting it on the body fixes the sticky-positioned button, but causes the content to be scrollable left-to-right (with a rubber-band effect, not really "scrollable" per se).

PR #711, opened 2018-12-02
fix(iOS): fix horizontal scroll, use fake sticky button on iOS

### LINKED ISSUE #667: iOS Safari allows scrolling left and right on the timeline
This is another regression introduced since the "document as scroller" fix. You can move you finger left and right, and the timeline moves horizontally, which is annoying.

nolanlawson:
fixes #667

To work around iOS issues, we're using `overflow-scroll: x` on the main content element, which fixes the horizontal scrolling issue (#667). However, because of #673, this causes the sticky button to no longer work, because the ancestor scroll element has changed. So for iOS only, we implement a faux-sticky button.

The downside is that the faux-sticky button looks a little jerky when it scrolls into view, but this is a small price to pay to fix the horizontal scrolling issue, and it only affects iOS anyway.


routes/_components/compose/ComposeBox.html
```diff
@@ -110,58 +79,48 @@ <h1 class="sr-only">Compose status</h1>
...
     oncreate () {
-      let { realm } = this.get()
-      if (realm === 'home') {
-        this.setupStickyObserver()
-      } else if (realm !== 'dialog') {
+      let { realm, replySpoiler, replyVisibility } = this.get()
+      if (realm !== 'home' && realm !== 'dialog') {
         // if this is a reply, populate the handle immediately
-        insertHandleForReply(realm)
+        /* no await */ insertHandleForReply(realm)
```
jokeyrhyme: Sorry, totally off topic...
I love this convention, I'm going to start doing this, too
(Kinda' wish it was an error call an asynchronous function without either `await` or `nowait`)

PR #720, opened 2018-12-04
chore(travis): use yarn instead of npm

nolanlawson:
Considering giving yarn a shot again, since I noticed that it takes ~90s for Travis to update the cache because npm seems to create a lot of dirty files on every npm install.


nolanlawson:
For whatever reason, yarn is dirtying the cache too, and anyway I have to manually install the latest version of yarn. This just doesn't seem worth it.

PR #721, opened 2018-12-04
chore(package): use events-lights instead of events

nolanlawson:
I don't see any reason not to switch. It's slightly smaller than the other events library and does everything we need.


nolanlawson:
3.45kb -> 1.27kb, why not

PR #727, opened 2018-12-04
Add the Pitch Black theme

sgenoud:
This theme is made with mobile OLED screens (iPhone XS for instance). This is the reason I used a pure black background.

capture d ecran 2018-12-04 17 11 35

capture d ecran 2018-12-04 17 13 15


nolanlawson:
Cool, I really like it! Thanks!

PR #733, opened 2018-12-05
WIP: detect private browsing and safari blocked cookies

### LINKED ISSUE #444: Doesn't work on iOS Safari with cookies disabled
Steps: disable cookies in Safari settings, navigate to pinafore.social.

Error is apparently:

    SecurityError: The operation is insecure.

AFAICT the issue isn't cookies per se, but the fact that Safari with "cookies" disabled also disables LocalStorage. Most likely there's no way to work around this, but opening a GH issue anyway to track it.

nolanlawson:
Attempt to fix #444, work in progress

If Pinafore is not going to work properly in private browsing mode, then I would at least like to show a warning. OTOH if there's something I can do to make it work in private browsing mode, then I'd like to do that as well.

Most browsers have an ephemeral storage for LocalStorage/IDB, so folks in private browsing should be able to "try before they buy." If not, I can try to sniff private browsing and at least offer a helpful error.

screenshot from 2018-12-05 00-19-28


nolanlawson:
Interestingly, both Firefox private and iOS Safari with "cookies blocked" disable IndexedDB.

nolanlawson:
screenshot from 2018-12-05 20-36-31

PR #747, opened 2018-12-07
Allow Media to be shown in a grid (as an option)

### LINKED ISSUE #258: Option for alternative UI for images/media
Currently all images are displayed below each other. A improvement would be to use the Halcyon grouping. Halcyon groups thumbnails in nice small group. The good points:

- Max height around 500px, even with multiple images
- [Focal point](https://medium.com/tootsuite/the-mastodon-spring-creators-release-65f60d0edb8c) used for thumbnails

Halcyon:
screenshot_1

Pinafore:
screenshot_2

sgenoud:
I have generally tried to keep how the non grouped behaviour was working. 

There is one main (intentional) change for the non grouped mode. Now, in the feed mode, images are centered accounting for the sidebar.

The grouped mode (optional, off by default) tries to show the images in a nice grid – it might result in some images not always cropped nicely. 
Nevertheless, it takes in account the focus that is set by the uploader.

Pretty
Pretty as well
Less pretty

Closes #258

### DIFF
--- a/routes/_components/AutoplayVideo.html
+++ b/routes/_components/AutoplayVideo.html
@@ -1,21 +1,67 @@
-<video
-  class="autoplay-video {className || ''}"
-  aria-label={ariaLabel || ''}
-  style="background-image: url({poster});"
-  {poster}
-  {width}
-  {height}
-  {src}
-  autoplay
-  muted
-  loop
-  webkit-playsinline
-  playsinline
-/>
+<div class="autoplay-wrapper  {$groupedImages ? 'fixed-size': ''}"
+     style="width: {width}px; height: {height}px;"
+>
+    <video
+    class="autoplay-video {$groupedImages ? 'fixed-size': ''}"
+    aria-label={ariaLabel || ''}
+    style="{focusStyle} background-image: url({poster}); "
+    {poster}
+    {width}
+    {height}
+    {src}
+    autoplay
+    muted
+    loop
+    webkit-playsinline
+    playsinline
+    />
+</div>
 <style>
+  .autoplay-wrapper {
+    position: relative;
+    margin: 0;
+    padding: 0;
+  }
+
   .autoplay-video {
     background-repeat: no-repeat;
     background-position: center;
-    background-size: contain;
+    background-size: cover;
+    display: flex;
+  }
+
+  .fixed-size {
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    overflow: hidden;
+  }
+
+
+  .fixed-size {
+    overflow: hidden;
+    width: 100%;
+    height: 100%;
   }
 </style>
+<script>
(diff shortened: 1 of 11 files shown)


sgenoud:
It was easy to do (as you can seen in the additional commit.

Do you have specific requirements for commit structures with PR (i.e. only one commit). I could not find anything in the contributing tab.

nolanlawson:
Nah, whatever commit format you want is fine. :) This looks great, and the test failures appear to be flaky tests. I'll keep re-running them. Thanks much!

routes/_components/AutoplayVideo.html
```diff
@@ -1,21 +1,67 @@
...
+
+  .fixed-size {
+    overflow: hidden;
+    width: 100%;
+    height: 100%;
+  }
+</style>
+<script>
+  import { store } from '../_store/store'
+  const coordsToPercent = coord => (1 + coord) / 2 * 100
```
nolanlawson: let's move coordsToPercent into its own file in `_utils`

routes/_components/dialog/components/ImageDialog.html
```diff
@@ -30,21 +36,26 @@
     max-height: calc(100% - 20px);
     overflow: hidden;
   }
+
+  .autoplay-video {
+    background-repeat: no-repeat;
+    background-position: center;
+    background-size: contain;
+  }
```
nolanlawson: We probably shouldn't repeat CSS selectors across multiple files, as I actually flatten all CSS for the production build (i.e. no per-component encapsulation) because it's smaller that way.

routes/_pages/settings/general.html
```diff
@@ -27,6 +27,11 @@ <h2>Accessibility</h2>
              bind:checked="$autoplayGifs" on:change="onChange(event)">
       <label for="choice-autoplay-gif">Autoplay GIFs</label>
     </div>
+    <div class="setting-group">
+      <input type="checkbox" id="choice-grouped-images"
+             bind:checked="$groupedImages" on:change="onChange(event)">
+      <label for="choice-groupes-images">Group images</label>
```
nolanlawson: This `label` doesn't match the `id`. Also I'd probably put this under "preferences" rather than "accessibility"

routes/_components/LazyImage.html
```diff
@@ -1,24 +1,40 @@
...
 </div>
 <style>
   .lazy-image {
+    margin: 0;
+    padding: 0;
     overflow: hidden;
+    display: flex;
   }
+
+  .fixed-size img {
```
nolanlawson: I would also prefer to add another class here (e.g. `fixed-size-img`) since it's a more performant selector

nolanlawson:
Actually don't sweat these changes; I'm happy to make them and then merge. Thanks again; this is awesome!!

PR #763, opened 2018-12-08
chore: use lighthouse-ci

### LINKED ISSUE #746: Run lighthouse tests in CI
https://github.com/ebidel/lighthouse-ci looks promising, and I don't want to regress on perf or a11y if we can avoid it.

nolanlawson:
fixes #746


nolanlawson:
I cannot seem to get lighthouse-ci to work. 

    Error from CI backend. invalid json response body at https://builder-dot-lighthouse-ci.appspot.com/ci reason: Unexpected token < in JSON at position 0

I also don't like that it still shows green even if this fails.

This would be neat to have, but it's just too hard to set up. I don't feel like setting up my own CI server for this right now.

Issue #27, opened 2018-04-09
Export as static site


nolanlawson:
This is useful for folks who want to self-host Pinafore but don't want to spin up a whole Node server to do so. It'd also be useful for hybrid mobile apps.

In theory `sapper export` should work, but I'm on an older version of Sapper and so it doesn't. Need to either 1) upgrade Sapper, or 2) write my own script to crawl the various URLs and generate a static site. (2) may be easier given all the modifications I've made to the Sapper code.

chr-1x:
you're probably pretty busy with higher priority stuff but i'd just like to throw in my support for this; i'd love to deploy it on cybre.space but i want it to be as light of an addition to the server as possible

nolanlawson:
Two issues it'd be hard to solve as a static site:

1. Dynamic URLs like `/tags/<myhashtag>` or `/settings/instances/<myinstance>` would not work when you refresh unless the Service Worker is activated, it's not a cold cache load, and it's not a hard-refresh. Probably a minor bug though.
2. More worryingly, CSP would have to be managed by the admin. Probably we could just provide an nginx/Apache snippet for folks to add, but CSP is important since we're injecting raw HTML into the DOM from instances (although your own instance would have to be malicious... but still).

nolanlawson:
This is probably blocked by #416 because latest Sapper has export built-in.

nolanlawson:
fixed by https://github.com/nolanlawson/pinafore/pull/775, there is now `npm run export`

PR #775, opened 2018-12-11
Update Sapper to latest

### LINKED ISSUE #416: Sync with latest Sapper

nolanlawson:
Fixes #416

I think we need to just bite the bullet and upgrade Sapper. Now that https://github.com/sveltejs/sapper/issues/422 is fixed, there's no big blocker to getting this done.

The main benefit is that we're no longer maintaining our own fork, and also switching columns is faster because it doesn't re-render the whole nav bar.

The downside AFAICT right now is:

1. mini-css-extract-plugin is not working correctly, fall back to whatever Sapper's default for CSS is, which is possibly not great for perf
2. need to set our own custom cache-control headers (or re-evaluate why I did this in the first place)
3. disable service worker on `localhost` but not `127.0.0.1` (this was a convenience during dev mode, maybe can just disable SW for dev mode)


nolanlawson:
for # 3 I just realized I can solve it by adding `127.0.0.1 lolcathost` to my `/etc/hosts` and then use that for devving

nolanlawson:
Total size of all JS chunks increases from 658KB to 720KB because of no `mini-css-extract-plugin` :disappointed:

nolanlawson:
Looks like this will also fix #27

nolanlawson:
Looks like there is a true bug in Sapper, and our tests caught it

nolanlawson:
Filed a bug on Sapper: https://github.com/sveltejs/sapper/issues/532

I took a quick glance and I'm not sure how to fix. So I may just change the structure of the URL.

nolanlawson:
https://pinafore-iccehsqqzl.now.sh/

nolanlawson:
I think it's time to bite the bullet on this one. Opened up a separate issue for `mini-css-extract-plugin` (#776), but everything else should be working.

PR #778, opened 2018-12-11
perf(svgs): remove unused svg <title>s

nolanlawson:
AFAICT these titles are not being read by screen readers, and anyway all of our usage of
SVGs is either `aria-hiden` or it has an `aria-label` somewhere around it. These `<title>`s
are not doing anything for us except taking up space.


nolanlawson:
Also removed the double `<svg style="display:none">` wrapper, not sure how that got in there but it's a mistake.

PR #784, opened 2018-12-12
Change thesize of sensitive media box

### LINKED ISSUE #771: in "grouped images" mode, the sensitive button should be the same size as media
It's all a fixed size, so it's weird to click that button and see the size change.

sgenoud:
It is now slightly bigger than what images in a box would look like.

I tried to make it fit exactly, but because we use a ratio of the width, we have subpixel computations error and the flow is redrawns slightly sometimes.

I also handle the case where you can have more the 4 images (even if it looks like mastodon does not allows it).

Closes #771

### DIFF
--- a/src/routes/_components/status/StatusMediaAttachments.html
+++ b/src/routes/_components/status/StatusMediaAttachments.html
@@ -1,6 +1,6 @@
 {#if sensitive }
 <div class={computedClass} style={customSize}>
-  <div class="inner-div">
+  <div class="status-sensitive-inner-div">
   {#if sensitiveShown}
   <button type="button"
           class="status-sensitive-media-button"
@@ -45,11 +45,11 @@
     background: none;
   }
 
-  .inner-div {
+  .status-sensitive-inner-div {
     height: 100%;
   }
 
-  .status-sensitive-media-container.grouped-images .inner-div {
+  .status-sensitive-media-container.grouped-images .status-sensitive-inner-div {
     position: absolute;
     width: 100%;
   }
@@ -92,7 +92,7 @@
     width: 100%;
     margin: 10px auto;
   }
-  .status-sensitive-media-container.status-sensitive-media-hidden:not(.grouped-images) {
+  .status-sensitive-media-container.status-sensitive-media-hidden.not-grouped-images {
       height: 200px;
   }
 
@@ -164,7 +164,7 @@
       computedClass: ({ sensitiveShown, $largeInlineMedia }) => classname(
         'status-sensitive-media-container',
         sensitiveShown ? 'status-sensitive-media-shown' : 'status-sensitive-media-hidden',
-        !$largeInlineMedia && 'grouped-images'
+        $largeInlineMedia ? 'not-grouped-images' : 'grouped-images'
       ),
       mediaAttachments: ({ originalStatus }) => originalStatus.media_attachments,
       sensitiveShown: ({ $sensitivesShown, uuid }) => !!$sensitivesShown[uuid],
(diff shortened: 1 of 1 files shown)


nolanlawson:
Awesome, sounds good! There seemed to be a minor lint error, BTW in the future you can fix those automatically with `npm run lint-fix` :smiley:

Also thanks a lot for this fix; even if the subpixels are slightly off, it's still less jarring to have it change size a little rather than a lot. I think this will be much nicer!

src/routes/_components/status/StatusMediaAttachments.html
```diff
@@ -43,8 +45,18 @@
     background: none;
   }
 
+  .inner-div {
```
nolanlawson: I disable Svelte's CSS encapsulation (for perf, to reduce the size of the CSS selectors), so in the future we probably want to avoid too-generic class names

src/routes/_components/status/StatusMediaAttachments.html
```diff
@@ -79,7 +91,9 @@
   .status-sensitive-media-container.status-sensitive-media-hidden {
     width: 100%;
     margin: 10px auto;
-    height: 200px;
+  }
+  .status-sensitive-media-container.status-sensitive-media-hidden:not(.grouped-images) {
```
nolanlawson: I tend not to use `:not` selectors (again, for perf). We can use a `.not-grouped-images` class here if we need to

sgenoud:
> Awesome, sounds good! There seemed to be a minor lint error, BTW in the future you can fix those automatically with npm run lint-fix 😃

Sorry about that. I tend to forget about linting (in my own project I have prettier set up to format everything on every save. I will try to not forget it!

> Also thanks a lot for this fix; even if the subpixels are slightly off, it's still less jarring to have it change size a little rather than a lot. I think this will be much nicer!

With the changes I have made, it does not change size at all. But there is a little bit of additional padding at the bottom of the screen.
The subpixels made everything move just enough for it to be annoying (to me at least).

I usually tend to value making my code simpler instead of more performant – but I think you tend to go more for perf.

nolanlawson:
No problem, I should probably set up prettier but I haven't really looked into it to know if it's something I'd like or not. :) I added an `.editorconfig` and StandardJS linting, but depending on your editor it may not set itself up automatically (I use WebStorm).

As for the class name and `:not` stuff, don't worry about it too much, it's mostly just my own little idiosyncracies. :D

Anyway this PR seems to have a slight bug; it looks like the images overflow when you click on them. I see this in both Chrome and Firefox.

screenshot from 2018-12-11 23-15-51

screenshot from 2018-12-11 23-15-58

sgenoud:
Yeah there is a bug, just saw it, about it fix it.

nolanlawson:
Restarted the flaky job. Once this is green I'll merge and probably do a release; I'm excited to get this out.

Also just tested it, and I can't even tell there is a subpixel issue (on desktop Chrome/Firefox anyway). Looks great!

sgenoud:
Thanks for your compliments (and thanks for all the rest of Pinafore 😉). 

I have been thinking of trying an alternative way of showing the images – in the same way than pixelfed does it (or Instagram, I think). Show one square image and a carousel. It might be a nice per instance option (i.e. if you have a more image oriented account you can see it in a different way). Just an idea so far.

nolanlawson:
Yeah, i would like to implement a carousel using scroll snap points, but I haven't gotten around to it.

I'm not sure what the deal with the tests is; it could be a true test failure. I'll look into it when I get a chance

sgenoud:
I had a quick look – it feels like the way I changed the structure of the list broke two tests. I might have a moment to fix it a bit later. We'll see who is going to be free first 😉

nolanlawson:
This seems to be some kind of issue with the target area for the toot changing size, which is subtly breaking some of the tests. I think clicking on the timestamp `<a>` may be less flaky, working on it...

nolanlawson:
Yeah, there seems to be an unrelated bug that this PR exposed somehow. Working on a fix: https://github.com/nolanlawson/pinafore/pull/790

PR #790, opened 2018-12-13
fix: fix flaky test by clicking relative date

nolanlawson:
This should fix test failures in #784 which somehow
exposed our flaky tests.


nolanlawson:
Yeah so there was a test that was just plain wrong. When you reply to a status and change the privacy of the reply, that change should manifest every time you reply to that same status, even in another timeline.

PR #794, opened 2018-12-13
fix: fix sensitive content warning on mobile

nolanlawson:
This fixes the padding and allows the warning to overflow.


nolanlawson:
Before and after:

screenshot from 2018-12-13 00-21-07

screenshot from 2018-12-13 00-20-39

PR #806, opened 2018-12-14
fix: logging out should result in still-visible homepage

### LINKED ISSUE #805: Logging out of only instance causes blank homepage
1. Log in to an instance
2. Log out
3. The home screen is a blank page

nolanlawson:
fixes #805


nolanlawson:
closing in favor of #807

Issue #179, opened 2018-04-19
Paste images from the clipboard to upload an image


nolanlawson:
Apparently the Mastodon web UI supports this: https://github.com/nolanlawson/pinafore/issues/65#issuecomment-381709685

PR #809, opened 2018-12-15
feat: add drag and drop for media uploads

### LINKED ISSUE #65: Drag/drop images onto the compose text box to upload

nolanlawson:
fixes #65

I also went ahead and fixed the media file upload types to match Mastodon (plus settings for the audio extension used by glitch.social).

The only thing I'm not 100% satisfied with is that you can only drag-drop over the compose text area with the textarea and the buttons, not, say, the whole page or the "toot!" button. I had to work around some issues with sticky-positioning on that button. But overall this is working, and I believe it's a niche feature anyway since it's not very discoverable.

This also adds in a web components polyfill, so if we want to start using webcomponents with more gusto, let's go for it


nolanlawson:
screenshot from 2018-12-15 01-47-20

spriteclad:
Will copy/paste image uploads be a thing too?

Fidgetcetera:
See #179, it's planned just not implemented yet.

nolanlawson:
Yep, copy-paste is not in there yet, sorry. This is just drag-and-drop.

PR #825, opened 2018-12-16
fix: calculate status aria-label in rIC

### LINKED ISSUE #752: Minimize cost of calculating aria-label
I'm thinking we could use innerText on the status content itself in order to get this, rather than using DomParser. But it may cause layout thrashing, so we have to be careful.

nolanlawson:
fixes #752

I think using `innerText` is too risky for this (since it could cause a style recalc), but we can at least move the `aria-label` calculations to a `requestIdleCallback`, where they're out of the critical rendering path.


nolanlawson:
It's too difficult to get tests to pass while also making this work, and it makes the code more fragile. Not worth the tradeoff I think

PR #826, opened 2018-12-16
fix: back button dismisses the modal dialog

nolanlawson:
Trying to implement back-button-respecting modals again (#60). This is not working yet

It seems like the root of my problems may be a Sapper issue actually. I can get hash-based modals to work correctly on every route except the `/` route.


nolanlawson:
I may have gotten this working using a hack on top of Sapper (https://github.com/nolanlawson/sapper/commit/cb21193fd1af5cc417a5ca9115b22d71a1f8c7fc), but now I'm wondering how to handle modals that open other modals, and it's not clear to me.

nolanlawson:
This is proving more difficult than I thought, because of Sapper insisting on re-rendering even when the route hasn't changed.

nolanlawson:
Hm, I still need to figure out how to handle nested modals, like the emoji dialog inside of the compose dialog.

nolanlawson:
This is almost working, but I'm really unhappy with all the hacks needed to handle modals that open other modals.

nolanlawson:
I have a strong suspicion that there are still bugs hiding in here, but let's give this a shot anyway.

PR #829, opened 2018-12-17
Fix Pleroma compability issue

### LINKED ISSUE #750: Pleroma issue with images
Hi there!

I'm trying to make Pinafore fully work with Pleroma, one of the issues is related with images as status attachments.

Pleroma doesn't seem to send metadata about attachments, this means there's no information about width or height of the images. And **pinafore uses this to set the image width and height**.

The result is: All images will show as width=300 height=200, which is the fallback size.

I understand a fallback size for images shown in the timelines (to prevent the scroll UI going back and forth during image loading), but... couldn't just use automatic size when opening the image inside the modal?

I know Pleroma compatibility is a "possible future goal", but this seems like a reasonable change for me and I could implement this in a PR if you're okay with it :)

This way, opening images in the modal would be full size, and improve the user experience for Pleroma users (and any other server technology for ActivityPub that doesn't send metadata about images).

Sirikon:
Should fix #750 and any issue with any Mastodon Client API-compliant server that doesn't provide attachment metadata (as it's optional).

### DIFF
--- a/src/routes/_components/dialog/components/ImageDialog.html
+++ b/src/routes/_components/dialog/components/ImageDialog.html
@@ -10,8 +10,7 @@
       class="image-modal-dialog-autoplay-video"
       aria-label="Animated GIF: {description || ''}"
       style="background-image: url({poster}); "
-      {width}
-      {height}
+      {style}
       {src}
       autoplay
       muted
@@ -22,8 +21,7 @@
   {:else}
     <img
       {src}
-      {width}
-      {height}
+      {style}
       alt={description || ''}
       title={description || ''}
     />
@@ -53,6 +51,11 @@
     oncreate,
     components: {
       ModalDialog
+    },
+    computed: {
+      style: ({ width, height }) => `
+        width: ${width ? width + 'px' : 'auto'};
+        height: ${height ? height + 'px' : 'auto'};`
     },
     methods: {
       show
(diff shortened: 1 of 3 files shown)


src/routes/_components/dialog/components/ImageDialog.html
```diff
@@ -10,8 +10,7 @@
       class="image-modal-dialog-autoplay-video"
       aria-label="Animated GIF: {description || ''}"
       style="background-image: url({poster}); "
-      {width}
-      {height}
+      {style}
```
nolanlawson: Having two style tags here may cause issues… can we move that `background-image: url({poster})` into the `computed` section?

Sirikon: Ups didn't saw that, sorry! Will fix right now.

Sirikon: Created a new computed property `videoStyle` and added the `background-image` just to this one (which just concats the `background-image` to the `style` computed property)

Sirikon:
For testing this, I've been working on this docker-compose definition for Pleroma: https://git.remolino.town/RemolinoTown/docker-pleroma

Also, can provide you credentials for a testing instance so you don't need to create one.

nolanlawson:
Thanks, that looks helpful! I think I have some throwaway accounts on some Pleroma instances anyway, I'll give them a shot…

nolanlawson:
Tested this out and it looks great. Thanks you!

nolanlawson:
If you want to test this out now, it should be live on https://dev.pinafore.social in a few minutes. :) Else it will arrive in the next release. Thanks again!

PR #836, opened 2018-12-18
feat: add full emoji picker using emoji-mart

### LINKED ISSUE #4: Full emoji keyboard
It's not ideal performance-wise, but not everyone has an emoji keyboard on their operating system or knows how to use it. We probably ought to have a full emoji keyboard beyond just the custom emoji.

nolanlawson:
Fixes #4 

screenshot from 2018-12-17 23-31-56

Normally I would love to write everything from scratch so that the dependencies can be as small as possible, but in the case of an emoji picker it would be a huge amount of work, and ultimately we can defer all the costs of the emoji picker until the user actually clicks the emoji button, so it's not _that_ bad to use a large dependency.

So anyway, this PR uses `emoji-mart`, which is a React component, so we can use `remount` which mounts a React component as a custom element. (We already have a custom element polyfill.) But to cut down on the size, I'm using `inferno-compat` instead of React. I'm also using the lightest version of `emoji-mart`, which uses native emoji (instead of a spritesheet) and `messenger.json` (which is the smallest emoji data set). Still, the overall size is pretty large:

screenshot_2018-12-17 webpack bundle analyzer 2

The `emoji-mart` chunk is 218kB, and there's also a chunk containing `inferno` and `inferno-compat` which is 35kB. Overall, this is increasing the size of all JavaScript assets from 762kB to 1020kB.

Still, an emoji picker is a pretty useful thing. Whether it's because the OS emoji picker is sub-par (Ubuntu, some Android keyboards) or because people just don't know that the OS emoji picker exists, it seems to be just expected for a chat app to bundle its own emoji picker. Using `emoji-mart` is about the least-bad option I could think of.

`emoji-mart` also gives us a lot out of the box, including custom emoji support and a "recently used" category based on LocalStorage. Also it's going to be familiar to Mastodon users.

TODO: emoji-mart does not have a dark mode, although I was able to style it partially based on the theme color.

### DIFF
--- a/bin/build-sass.js
+++ b/bin/build-sass.js
@@ -4,9 +4,11 @@ import sass from 'node-sass'
 import path from 'path'
 import fs from 'fs'
 import pify from 'pify'
+import CleanCSS from 'clean-css'
 
 const writeFile = pify(fs.writeFile.bind(fs))
 const readdir = pify(fs.readdir.bind(fs))
+const readFile = pify(fs.readFile.bind(fs))
 const render = pify(sass.render.bind(sass))
 
 const globalScss = path.join(__dirname, '../scss/global.scss')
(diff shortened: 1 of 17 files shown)


nolanlawson:
Just as a sanity check, I compared inferno to vanilla React. The React chunk is 103kB, the Inferno one is 35kB. So yeah, it's worth it.

PR #843, opened 2018-12-19
fix(picker): only focus picker search on desktop

nolanlawson:
I find it jarring if the emoji picker dialog automatically pops open the soft keyboard; it causes two extra layouts and can cause part of the dialog to be hidden. We can't detect the existence of a soft vs hard keyboard, but we can sniff for it by checking for the mobile size.


nolanlawson:
On second thought, a much better test is whether the user has touched the screen or not (i.e. it's a touchscreen). This will also capture iPad, and it won't capture very narrow desktop views (such as Mastodon as a Sidebar Firefox add-on).

PR #847, opened 2018-12-19
Mention Serveo as an alternative to setting up Mastodon

sorin-davidoi:


nolanlawson:
Thanks for the PR! Is it necessary to proxy to localhost to test Pinafore, though? I often connect my local dev Pinafore directly to malfunctioning.technology. Arguably you can't run the TestCafe tests that way, but the TestCafe tests should fail anyway, because they assume certain canned data and accounts anyway.

nolanlawson:
Closing because I'm not 100% sure this is necessary, although we could definitely use a better "quickstart" for setting up the local Mastodon server. I did add some new info to CONTRIBUTING.md to try to help, though.

PR #854, opened 2018-12-23
fix: import directly from lodash-es

nolanlawson:
This format is terser and also reduces the total bundle size by 1kB.


nolanlawson:
Turns out this breaks the server for some reason, and I'd rather not run the server with `-r esm` to fix it.

PR #855, opened 2018-12-23
perf: use webpack splitChunks defaults

nolanlawson:
I don't think there's much reason to override these defaults. This actually reduces the non-logged-in homepage JS bundle size by 5kB, although increasing the size of all chunks by 150kB. Doesn't seem to make much sense though to think we can beat the Webpack defaults, probably better to just use the defaults here.

### DIFF
--- a/webpack/client.config.js
+++ b/webpack/client.config.js
@@ -35,10 +35,6 @@ module.exports = {
       terser()
     ],
     splitChunks: {
-      chunks: 'async',
-      minSize: 5000,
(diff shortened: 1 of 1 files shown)


nolanlawson:
Also this reduces the total number of webpack files (79 -> 63). Surprisingly though it causes a worse compression ratio: total 859034 -> 1127741, gzip -9 total 257072 -> 314786, compression ratio 29.9% -> 27.9%. Still, I'm happy just to stick with the defaults.

Issue #848, opened 2018-12-19
split light/dark themes in the theme selector


Fidgetcetera:
There seems to be an implicit but unlabeled split between the themes in the theme picker as light on the left column, dark on the right. it may be worth explicitly labeling them with a "light background" and "dark background" indicator at the top, just to be super clear. 

the outlines around the primary color that indicate the background color aren't very visible against their own corresponding themes which could cause confusion, and labels are just nice.

nolanlawson:
I agree, this could be labeled and that would clear up some confusion.

nspenner:
I'm not sure about Nolan's thoughts on this issue, but I was able to add light/dark labels relatively easily because each column already has a `dark` property attached:

![theme_labels](https://user-images.githubusercontent.com/14966249/50377800-61edbd00-05e9-11e9-831d-a4b85bf98404.gif)

Is that what you were thinking? I'll wait to open or push anything pending input from Nolan.

nspenner:
Oh, we just commented at exactly the same time! I will open a branch with these changes then.

nolanlawson:
fixed in #867!

PR #867, opened 2019-01-04
Add labels to theme groups

nspenner:
Adds a label to each theme group (light vs dark) as referenced in #848:

theme_labels


nolanlawson:
Awesome, thanks!

PR #870, opened 2019-01-08
Add support for keyboard shortcuts

szermatt:
Hi!

I gave adding keyboard shortcut a try, since it's the single feature I missed from pinafore.If you're interested in merging, please have a look and let me know what you'd like to change or fix.

Thanks,

  Stephane


src/routes/_components/DynamicPageBanner.html
```diff
@@ -10,8 +10,9 @@ <h1 class="dynamic-page-title" aria-label={ariaTitle}>{title}</h1>
   <button type="button"
           class="dynamic-page-go-back"
           aria-label="Go back"
-          on:click="onGoBack(event)">Back</button>
+          on:click|preventDefault="onGoBack()">Back</button>
```
nolanlawson: This is a cool Svelte feature that I wasn't aware of, but in the future maybe it should go in another PR? :)

src/routes/_utils/scrollIntoView.js
```diff
@@ -0,0 +1,64 @@
+import {
+  getScrollContainer,
+  getOffsetHeight } from './scrollContainer'
+import { smoothScroll } from './smoothScroll'
+
+function getTopOverlay () {
+  return document.getElementById('main-nav').clientHeight
```
nolanlawson: This element can be cached

nolanlawson:
Wow, this is a really impressive PR. I looked it over and everything looks amazing, including the new tests! I also like that you import the keyboard shortcut logic asynchronously, even if it's probably a minor perf improvement.

I'm afraid I bitrotted this by merging some other conflicting PRs, but I can fix the conflicts and then when the tests pass, I'll give it another quick test, but this looks very good overall! Thanks a million!

src/routes/_utils/scrollIntoView.js
```diff
@@ -0,0 +1,64 @@
...
+  if (!element) {
+    return false
+  }
+  let rect = element.getBoundingClientRect()
+  let offsetHeight = getOffsetHeight()
+  let topOverlay = getTopOverlay()
+  return rect.top < offsetHeight && rect.bottom >= topOverlay
+}
+
+export function firstVisibleElementIndex (items, itemElementFunction) {
```
nolanlawson: Wondering if we could do a binary search here instead of checking each element one at a time, but that may be an excessive micro-optimization.

nolanlawson:
I played around with this, and everything looks amazing. My only minor feedback:

- `?` doesn't seem to work for me, at least in Firefox and Chrome on Linux. `h` works though
- Mastodon has a nice shortcut where `1` is the first column, `2` is the second, etc. I think we could do that too for nav items
- Discoverability is a bit low, maybe we should put a link to the dialog in the settings menu or something?

In any case I think this looks great, and I'm happy to get it in. Thanks for implementing this!

nolanlawson:
Don't worry about all these follow-ups - I've fixed them in additional PRs. :) You really went above and beyond in this PR - love the unit tests as well!

szermatt:
Ok. Thank you for merging ... and fixing ! I would have been happy to add any follow-ups but it would have taken me longer :) I'll send proposals for the remaining comments next week-end.

src/routes/_components/DynamicPageBanner.html
szermatt: Ok. I didn't realize I was adding something new... all I wanted was to avoid having to define two methods.

src/routes/_utils/scrollIntoView.js
szermatt: Yes. I'll send out a fix.

src/routes/_utils/scrollIntoView.js
szermatt: This is possible. I'll give it a try to see how much it improves things. Whether it'll be worth it depends on how big the array of visible items can grow on virtual lists.

szermatt: I tried to address this in #910. Let's continue the conversation there.

src/routes/_utils/scrollIntoView.js
szermatt: I tried to address it in #910. Let's continue the conversation there.

Issue #35, opened 2018-04-09
Self-hosted as an alternative frontend, in "single-instance mode"


nolanlawson:
This would be useful for admins who just want Pinafore as an alternative UI that only works for their instance. It's possible, but would probably need a separate build system in order to lock it to a single instance and automatically log folks in.

nolanlawson:
I think this could be doable, but Pinafore would have to integrate with the Mastodon/Pleroma server on the backend in order to share the cookie. E.g. it could be embedded into the HTML or accessible via a REST endpoint or something.

nolanlawson:
Another nice feature would be `DEFAULT_LIGHT_THEME=foo` or maybe just `DEFAULT_THEME` or something. An instance like toot.cafe would probably want Majesty for the light theme and Sorcery for the dark theme.

nolanlawson:
I've been thinking more about this, in context of #891. I think if instance admins wanted to host Pinafore at `example.com/pinafore`, then that could work. It would solve the cookie issue, because it'd be same-origin. There is also a precedent for this in that Pleroma hosts the Mastodon FE at `/web`.

I envision something like:

    PINAFORE_INSTANCE=example.com PINAFORE_PATH=pinafore yarn build && node server.js

If `PINAFORE_PATH` is omitted, then the default path is used. This would cover cases like #891 where they want Pinafore to be the primary frontend even for logged-out users.

I believe Sapper supports paths, but I'd need to check.

The logged-out homepage could also contain some boilerplate saying "You're not logged in; please log in at example.com".

nolanlawson:
I would also ideally like the instance-switcher to be completely disabled in this case, to truly "lock" Pinafore to a given instance.

Also `auth()` would have to be overridden to use the cookie rather than a token in localStorage.

uiri:
How would the environment variables translate for a static site export?
Right now this is the easiest way to deploy pinafore as a primary/secondary frontend onto an instance, at least in the case of pleroma.

Outside of that concern, the instance and path configuration model seems good to me.

The ordinary Pinafore login flow is fine. I think it should be preserved other than skipping the instance selection screen. I think that would avoid the `auth()` issue.

A way to disable the instance-switcher would be nice. Perhaps that should be a separate setting from `PINAFORE_INSTANCE` and `PINAFORE_PATH`?

nolanlawson:
> How would the environment variables translate for a static site export?

[...]

PR #872, opened 2019-01-10
Fix #869: Allow logged out user to view content on default instance

### LINKED ISSUE #869: Allow logged out user to view content from a default instance
I suppose this is related to #35 -- I'm not concerned about the login flow, but right now there is no obvious way to use Pinafore as an alternative frontend and still allow for canonical links to an instance's toots and user profiles.

There are definitely APIs for this since both Mastodon's frontend and Pleroma's frontend allow for this but I suspect that they differ between Mastodon and Pleroma...

uiri:
This is at the works-on-my-server stage. It still needs tests.

Open questions:
- Is a variable buried deep inside the source the right place for configuring the default instance?
- Should support for no default instance be maintained? If not, what should the default default instance be?

For now, I've left the default default instance as `mastodon.social` even though obviously I use a different value for my testing.

### DIFF
--- a/src/routes/_actions/accounts.js
+++ b/src/routes/_actions/accounts.js
@@ -1,3 +1,4 @@
+import { registerDefaultInstance } from './addInstance'
 import { getAccount } from '../_api/user'
 import { getRelationship } from '../_api/relationships'
 import { database } from '../_database/database'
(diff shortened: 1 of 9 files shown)


nolanlawson:
This is an interesting PR, thanks! I'm not sure I'd like to merge it, though, since I kind of disagree that Pinafore should privilege any one instance over another one (e.g. `mastodon.social`). I think it should be a neutral third-party app (unless maybe when self-hosted by another instance, TBD).

If you want to override `currentInstance` then unfortunately it's going to be a bit hard because it's referenced everywhere. The best solution may be to turn it into a computed property (e.g. [here](https://github.com/nolanlawson/pinafore/blob/master/src/routes/_store/computations/instanceComputations.js) and then make sure that `currentInstance` is never `set()` but rather computed from some other property. Otherwise you will probably miss a spot.

nolanlawson:
Closing for now, as I'm not sure about having a default instance for all of Pinafore. Maybe in order to solve #35, we could instead have a global environment variable that could be used during building, and which would auto-log in to the given instance by default using a cookie.

PR #873, opened 2019-01-10
Wrap LONG posts with spoilers

uiri:
Some ActivityPub software facilitates long form blog posts being pushed
out onto timelines.
Some ActivityPub software have toot lengths which are much larger than
Mastodon's default of 500.

This wraps spoiler tags around those statuses.

### DIFF
--- a/src/routes/_components/status/Status.html
+++ b/src/routes/_components/status/Status.html
@@ -116,6 +116,8 @@
   import { getAccessibleLabelForStatus } from '../../_a11y/getAccessibleLabelForStatus'
   import { formatTimeagoDate } from '../../_intl/formatTimeagoDate'
 
+  const LONG_POST_LENGTH = 1024
+  const LONG_POST_TEXT = 'Long'
   const INPUT_TAGS = new Set(['a', 'button', 'input', 'textarea'])
   const isUserInputElement = node => INPUT_TAGS.has(node.localName)
   const isToolbar = node => node.classList.contains('status-toolbar')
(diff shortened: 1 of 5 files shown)


src/routes/_components/status/Status.html
```diff
@@ -194,7 +194,7 @@
       originalAccount: ({ originalStatus }) => originalStatus.account,
       originalAccountId: ({ originalAccount }) => originalAccount.id,
       visibility: ({ originalStatus }) => originalStatus.visibility,
-      spoilerText: ({ originalStatus }) => originalStatus.spoiler_text,
+      spoilerText: ({ originalStatus }) => originalStatus.spoiler_text || ((originalStatus.content || '').length > 1024 ? "LONG" : undefined),
```
nolanlawson: I would like both `1024` and `"LONG"` to be pulled out as constants somewhere. Also wondering if instead of "LONG" it could show the number of characters or something.

nolanlawson: On second thought, let's just make it "Long"

nolanlawson:
Nice PR! As a tip, in the future you can run `npm run lint-fix` to fix lint errors. I'll merge this soon, LGTM. Thanks much!

uiri:
Oops, sorry, I guess I forgot to run the lint checker before sending out the PR.

I'll add it as a commit hook on my dev machine.

nolanlawson:
It looks like the test is failing because there's a post that's not "really" 1024 characters, but it counts as such because we're not considering that URLs are automatically shortened. We should use the same length-calculation system that Mastodon itself uses (or at least as the 500-char limit uses).

nolanlawson:
Tested manually to make sure it works; there's currently no way to test using our local Mastodon server because it cuts off posts at 500 chars. Thank you for the PR!

PR #891, opened 2019-01-13
Fix #869: Allow logged out user to view content on default instance

### LINKED ISSUE #869: Allow logged out user to view content from a default instance

uiri:
(Re-opening #872 since the stated reason for closure was addressed before it was closed).

I guess this still needs tests. The configuration variable for the default instance lives in `src/route/_static/config.js`. I'm not sure if that's the right place for it. The default value for it is "`false`" which maintains the existing experience for logged out users. This avoids Pinafore even unintentionally privileging a certain instance by making it the default default instance.

### DIFF
--- a/src/routes/_actions/accounts.js
+++ b/src/routes/_actions/accounts.js
@@ -1,3 +1,4 @@
+import { registerDefaultInstance } from './addInstance'
 import { getAccount } from '../_api/user'
 import { getRelationship } from '../_api/relationships'
 import { database } from '../_database/database'
@@ -58,6 +59,9 @@ export async function clearProfileAndRelationship () {
 
 export async function updateProfileAndRelationship (accountId) {
   let { currentInstance, accessToken } = store.get()
+  if (currentInstance === undefined || currentInstance === null) {
+    currentInstance = await registerDefaultInstance()
+  }
 
   await Promise.all([
     _updateAccount(accountId, currentInstance, accessToken),
@@ -67,6 +71,9 @@ export async function updateProfileAndRelationship (accountId) {
 
 export async function updateRelationship (accountId) {
   let { currentInstance, accessToken } = store.get()
+  if (currentInstance === undefined || currentInstance === null) {
+    currentInstance = await registerDefaultInstance()
+  }
 
   await _updateRelationship(accountId, currentInstance, accessToken)
 }
--- a/src/routes/_actions/timeline.js
+++ b/src/routes/_actions/timeline.js
@@ -1,3 +1,4 @@
+import { registerDefaultInstance } from './addInstance'
 import { store } from '../_store/store'
 import { getTimeline } from '../_api/timelines'
 import { toast } from '../_components/toast/toast'
@@ -89,6 +90,9 @@ async function fetchTimelineItemsAndPossiblyFallBack () {
     lastTimelineItemId,
     online
   } = store.get()
+  if (currentInstance === undefined || currentInstance === null) {
+    currentInstance = await registerDefaultInstance()
+  }
 
   let { items, stale } = await fetchTimelineItems(currentInstance, accessToken, currentTimeline, lastTimelineItemId, online)
   addTimelineItems(currentInstance, currentTimeline, items, stale)
--- a/src/routes/_api/utils.js
+++ b/src/routes/_api/utils.js
@@ -10,6 +10,10 @@ export function basename (instanceName) {
 }
 
 export function auth (accessToken) {
+  if (accessToken === undefined || accessToken === null) {
+    return {}
+  }
+
   return {
     'Authorization': `Bearer ${accessToken}`
   }
--- a/src/routes/_components/profile/AccountProfile.html
+++ b/src/routes/_components/profile/AccountProfile.html
@@ -7,7 +7,9 @@
   <div class="account-profile-grid-wrapper">
     <div class="account-profile-grid">
       <AccountProfileHeader {account} {relationship} {verifyCredentials} />
+      {#if $isUserLoggedIn}
       <AccountProfileFollow {account} {relationship} {verifyCredentials} />
+      {/if}
       <AccountProfileNote {account} />
       <AccountProfileMeta {account} />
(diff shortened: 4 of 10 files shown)


nolanlawson:
Just so I understand: your goal for this to fix #35 so that an instance admin of `myinstance.com` could host Pinafore at e.g. `pinafore.myinstance.com`, and the default instance would be `myinstance.com`, correct? Am I understanding your goal here?

I think in that case the best place to put this kind of config may be in an environment variable, e.g. `DEFAULT_INSTANCE=myinstance.com`. If it's set, then the app would use that as the default instance, but if not, then Pinafore would behave like normal. Then we would also need instructions in either Admin-Guide.md or CONTRIBUTING.md so that folks who self-host would know how to enable this feature if they wanted to. You may also need to put this env variable in the webpack config.

src/routes/_actions/accounts.js
```diff
@@ -67,6 +71,9 @@ export async function updateProfileAndRelationship (accountId) {
 
 export async function updateRelationship (accountId) {
   let { currentInstance, accessToken } = store.get()
+  if (currentInstance === undefined || currentInstance === null) {
+    currentInstance = await registerDefaultInstance()
```
nolanlawson: this doesn't need to be async AFAICT

src/routes/_actions/timeline.js
```diff
@@ -88,6 +89,9 @@ async function fetchTimelineItemsAndPossiblyFallBack () {
     lastTimelineItemId,
     online
   } = store.get()
+  if (currentInstance === undefined || currentInstance === null) {
+    currentInstance = await registerDefaultInstance()
+  }
```
nolanlawson: It makes me a bit uncomfortable that we need to put these checks in so many different places. I imagine we must be missing a few (e.g. the search page). Can we not put a string, `defaultInstance` into `store.js` and look for that instead?

src/routes/_api/utils.js
```diff
@@ -10,6 +10,10 @@ export function basename (instanceName) {
 }
 
 export function auth (accessToken) {
+  if (accessToken === undefined || accessToken === null) {
+    return {}
+  }
+
```
nolanlawson: I would rather we fix the places where the accessToken is null. Usually I'm pretty specific about not including the accessToken in cases where it's not needed (e.g. `/api/v1/instance`).

src/routes/_components/profile/AccountProfile.html
```diff
@@ -4,7 +4,9 @@ <h1 class="sr-only">Profile for {accountName}</h1>
   <div class="account-profile-grid-wrapper">
     <div class="account-profile-grid">
       <AccountProfileHeader {account} {relationship} {verifyCredentials} />
+      {#if $isUserLoggedIn}
       <AccountProfileFollow {account} {relationship} {verifyCredentials} />
+      {/if}
```
nolanlawson: So users can browse other users' profiles even if they're not logged in?

src/routes/_components/status/Status.html
```diff
@@ -31,7 +31,7 @@
     <StatusDetails {...params} />
   {/if}
   <StatusToolbar {...params} on:recalculateHeight />
-  {#if replyShown}
+  {#if replyShown && $isUserLoggedIn}
```
nolanlawson: Ideally we would disable the reply button if the user isn't logged in, or (even better) make it so that users can't browse timelines unless they're logged in.

nolanlawson:
Is your goal to show the default public federated/local timeline even if the user isn't logged in? I didn't realize that until I started looking at your code. I admit it's a use case I hadn't even considered...

uiri:
To clarify the use case which I wish to address here:

1. Pinafore is hosted as the front-end for an instance (let's call it `example.social`).
2. A fediverse user on a different instance, seeing a toot propagated from `example.social` clicks through to `example.social` to view the toot on that instance, or to view the profile of a user on `example.social` on that instance.

The desired behaviour is that the toot or profile shows up to the fediverse user even though they are not logged into the Pinafore instance of `example.social`. Relatedly, since the URLs rely on toot ids and account ids, the links on `example.social` will point to different toots or account profiles even if that fediverse user uses the Pinafore instance to log into their account on their home instance (i.e. which is not `example.social`).

The other use case is that, a user on `example.social` shares a link to a toot via other means with someone who may or may not have a fediverse account but is nonetheless not logged in to `example.social`.

The actual behaviour is that Pinafore says that a status thread or account profile will be loaded only after a user logs in.

src/routes/_components/profile/AccountProfile.html
uiri: Exactly :100:

src/routes/_api/utils.js
uiri: The accessToken may be needed to view toots that have privacy settings (e.g. DMs, follower-only).

I am not sure if it is *really* needed for viewing account profiles though. I think not, but maybe it flows through to the timeline which would then need it.

src/routes/_static/config.js
```diff
@@ -0,0 +1 @@
+export const DEFAULT_INSTANCE_HOSTNAME = false
```
nolanlawson: `process.env.DEFAULT_INSTANCE_HOSTNAME` and inserting via Webpack would make it easier for instance admins to change this value

nolanlawson:
I have some thoughts about how we could make this work in https://github.com/nolanlawson/pinafore/issues/35#issuecomment-470180217, seems like we could use this PR as a starting point and solve #35 entirely?

nolanlawson:
Closing for now, I think this PR is interesting but it's a bit out-of-scope for Pinafore, especially given that we still don't have the option to run Pinafore as an alternative frontend (at least, a blessed way of doing it). Thanks for the PR though!

PR #892, opened 2019-01-13
Make external links open in new tab

uiri:
(Reopening #543 since the bug still persists for Pleroma instances)

External links inside toots (when using a Pleroma instance) opens up in a new tab (using target='_blank') instead of the same tab.


nolanlawson:
Thanks!

PR #910, opened 2019-01-21
Follow-up for pull request  #870

szermatt:
Hi !

I tried to address the remaining open comments on #870:

caching the main-nav element, in 59b13dc

I tried using binary search for firstVisibleElementIndex, but the result was actually slower than the straightforward implementation on Chrome :)  Virtual lists seem to have at most 20 items, so the size is limited.

However, while doing it, I noticed that the return value of firstVisibleElementIndex was wrong in the case where there was no visible elements, so I fixed it in b16406f

### DIFF
--- a/src/routes/_utils/scrollIntoView.js
+++ b/src/routes/_utils/scrollIntoView.js
@@ -41,7 +41,7 @@ export function firstVisibleElementIndex (items, itemElementFunction) {
       break
     }
   }
-  return {first, firstComplete}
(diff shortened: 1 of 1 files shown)


nolanlawson:
Awesome, thanks, looks good to me! I agree that for 20 items, binary search is overkill. Sorry for sending you down a rabbit hole!

The caching looks good and the new loop logic looks great too. I'll run `npm run lint-fix` and then merge. :)

PR #919, opened 2019-01-26
feat: add option to underline links in toot text

nolanlawson:
Disabled by default, as we already have good color contrast in the links.


nolanlawson:
screenshot from 2019-01-26 12-27-36

screenshot from 2019-01-26 12-29-58

screenshot from 2019-01-26 12-30-09

Hm, not sure if it makes sense to also underline the names and dates. Those are kind of clear from context, I think.

nolanlawson:
Nah, if people want underlines, I'll enable underlines everywhere.

PR #921, opened 2019-01-27
chore: update several deps

nolanlawson:


nolanlawson:
Looks like Webpack broke something. I'll have to test each dep independently.

PR #926, opened 2019-01-27
Update webpack

nolanlawson:


nolanlawson:
Seems I hit https://github.com/webpack/webpack/issues/8656

nolanlawson:
Based on https://github.com/webpack/webpack/issues/8656#issuecomment-456713191 looks like me way need to switch to yarn. I've tried a few things, and this is the only one that's working.

PR #930, opened 2019-02-03
Add shortcuts to the media dialog

charlag:
Hey
I have no idea what I'm doing but seems to be working and I really want this :sweat_smile:

### DIFF
--- a/src/routes/_components/dialog/components/MediaDialog.html
+++ b/src/routes/_components/dialog/components/MediaDialog.html
@@ -50,6 +50,9 @@
 
 
 </ModalDialog>
+
+<Shortcut scope='mediaDialog' key="ArrowLeft" on:pressed="prev()" />
+<Shortcut scope='mediaDialog' key="ArrowRight" on:pressed="next()" />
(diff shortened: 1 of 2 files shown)


charlag:
Oops, sorry for the issues. I've fix styling issues and force-pushed but it doesn't seem to be picked up by CI?..

nolanlawson:
I'll rebase and give this another go. I'm in the middle of a big rewrite on the MediaDialog, but I see what you're trying to do and it LGTM. :) Thanks!

src/routes/_components/dialog/components/MediaDialog.html
```diff
@@ -198,6 +209,18 @@
           this.scrollToItem(i, true)
         }
       },
+      next () {
+        let { scrolledItem } = this.get()
+        if (scrolledItem < 4) {
```
nolanlawson: this should be `length` instead of 4, since there may be less than 4 media.

nolanlawson:
Thanks again! :)

PR #939, opened 2019-02-03
fix: use react-lite instead of inferno-compat

nolanlawson:
react-lite is smaller than inferno-compat (35.6kB vs 28.1kB) and appears to work the same.


nolanlawson:
On second thought, the tradeoff doesn't seem worth it, let's stick with the devil we know

PR #941, opened 2019-02-05
fix: use preact-compat instead of inferno-compat

nolanlawson:
Compared to #939, this really gives us a perf win to write home about.

| package | chunk size|
| ---- | --- |
| inferno-compat | 35.36k |
| react-lite | 28.06k |
| preact-compat | 18.88k |

There was a small modification necessary for emoji-mart to be preact-compatible, though: https://github.com/missive/emoji-mart/pull/268


nolanlawson:
Nevermind, there are bugs. I cannot get preact to work correctly.

nolanlawson:
In particular, the search text is kept even after closing and reopening the modal, and even after I hack it by clearing the search, there are bugs where it doesn't show any emoji and you can't search at all. So I'm not sure what to do; I guess preact is just not in the cards.

PR #973, opened 2019-02-13
perf: reduce server-side memory usage of Pinafore

### LINKED ISSUE #971: Server-side memory usage reported as "twice the size of Pleroma"
That seems out of whack; there may be a way to reduce server-side memory usage.

nolanlawson:
fixes #971 

This PR does a few things:

- remove `cross-env` from the `start` script, as I can pretty safely say there are no admins running Pinafore in production on Windows
- directly run `node` in Docker instead of running either `yarn start` or `npm start`
- change the README to recommend admins to run `node` directly

This doesn't change the way Pinafore runs in Zeit (where it will still run `npm start` or `yarn start` presumably), but at least for self-hosters this should reduce the memory usage dramatically (300kB -> 48kB if the sum of RSS in `ps aux` is to be believed).


sorin-davidoi:
> 300kB -> 48kB

Wow, was this mostly caused by `cross-env` or running via `npm start`?

sorin-davidoi:
Nevermind, just read the issue! Was aware that the overhead of these things is so big.

nolanlawson:
Just realized this is actually 300MB -> 48MB, so the win is actually much bigger :smiley:

Issue #985, opened 2019-02-14
Zeit CDN not working


nolanlawson:
For some reason, Zeit's CDN seems to suddenly not be working. This results in an F score from WebPageTest on First Byte Time and an X on Effective Use of CDN. https://www.webpagetest.org/result/190214_Z9_86475ea42d9e2d3b3220084746eda5e2/

This used to work, not sure what happened.

nolanlawson:
I wonder if it's because of the nonce in the CSP header? Maybe Zeit now uses that as a clue to disable the CDN. I could use checksums instead.

nolanlawson:
Repro: not seeing any CDN headers when running this command:

    curl -sLv pinafore.social/favicon.png -o /dev/null

nolanlawson:
According to [this post](https://community.cloudflare.com/t/content-security-policy-csp-cloudflare-generated-nonce/6131) it does seem like CSP nonce may cause an issue.

nolanlawson:
Supposedly all the `*.now.sh` URLs are supposed to be CDN-powered if you're on the paid plan, but this doesn't seem to be happening. I'm assuming it's because I'm on the legacy paid plan rather than the new one. Seems weird that that would cause CDN to just stop working though.

nolanlawson:
Upgraded to now v2, switched to the new Unlimited plan, turned the CDN off and on just to be safe, and it's _still_ not using Cloudflare. Looks like I'm going to need to ping support.

nolanlawson:
Opened an issue with Zeit: https://spectrum.chat/zeit/general/cdn-not-working~9d8d0393-112f-4c12-a49e-6364be9276d8

nolanlawson:
Boy, this is frustrating. It's a static site!
![screenshot from 2019-02-16 16-42-45](https://user-images.githubusercontent.com/283842/52906836-f6e7ed80-3209-11e9-9d7d-49f76e0dfedf.png)

nolanlawson:
OK so Zeit has acknowledged it, and it's apparently a global outage. So it's not my bug at all. https://zeit-status.co/incidents/2019/01/28/upstream-dns-issues/

Although in retrospect I'm glad I went to all this work to get Pinafore-as-a-static-site to work, because it fixed some bugs in exporting and also makes me more confident I can make the server as fast as possible.

nolanlawson:
The CDN works in Europe:

[...]

PR #988, opened 2019-02-15
fix: increase cache, use csp checksums over nonce

nolanlawson:
attempt to address #985

This is a somewhat hacky solution to the problem. I'm trying to avoid nonces in favor of checksums so that (presumably) Zeit's CDN (Cloudflare I think?) can cache the content more reliable.

TODO: core HTML files are still being served with cache-control max-age=600, which I believe is hard-coded in Sapper somewhere


nolanlawson:
I'm not a huge fan of this fix, but it's worth at least trying out to see if it helps with the caching issues.

PR #993, opened 2019-02-15
fix: use "public" for cache-control

nolanlawson:
another attempt to address #985

Based on the Cloudflare docs, it seems "public" may convince it to cache it.


nolanlawson:
This really doesn't seem to be working in terms of Zeit/Cloudflare, but it seems worth adding anyway to make it clear to proxies that all assets are public.

PR #998, opened 2019-02-16
chore: upgrade to zeit 2

nolanlawson:
More work on fixing #985.

I'm not a huge fan of this implementation, because we now basically have two different "production" implemenations - one using `node server.js` (using Express + Sapper middleware), and another using Sapper export plus a static fileserver in Now.

But it does work, and the difference between the two is minimal. The only thing I had to do slightly differently was handling dynamic URLs like `/accounts/<accountId>` - instead of letting Sapper middleware handle it, I just return `service-worker-index.html`. This is fine, because literally the only case where this matters is if a user is on one of those dynamic pages, and either 1) hard-refreshes, and 2) doesn't have a Service Worker, and refreshes. Those cases are so rare that the difference doesn't matter much.

More troubling is having to manually manage this entire CSP mess, as well as a big regex to identify the dynamic URLs. But since we have a dev server to catch any mistakes I might make, I'm not too concerned. But it could be improved.


nolanlawson:
Also once https://github.com/sveltejs/sapper/pull/566 is merged into Sapper, we can potentially just use `sapper export` as the "normal" build process. The reason I don't want to do that now is because without that PR, Sapper would always use port 3000 for exporting, which may conflict with self-hosters who might be running something else on 3000, e.g. Mastodon.

nolanlawson:
Just occurred to me that we will lose the link rel=preload in the header due to this PR, but I think it's worth it to get the CDN working and switch to a faster static model. Even without the CDN, this seems faster since it doesn't have to go through a Node server.

Rich-Harris:
> Just occurred to me that we will lose the link rel=preload in the header due to this PR

keep an eye on this issue 😀 https://github.com/sveltejs/sapper/issues/402

Rich-Harris:
oh man, I just saw https://github.com/sveltejs/sapper/pull/568 🤦‍♂️

PR #1028, opened 2019-02-22
fix: j/k shortcuts correctly set active element

### LINKED ISSUE #1018: Keyboard shortcuts do not actually change focus
>J and K, when moving between toots, does not yet actually focus the new toot, only styles it differently from the old one? If I turn off virtual cursor with NVDA, J and K don't speak or move DOM focus.

I took a look at the Mastodon implementation, and it correctly changes the focus as you press J and K.

nolanlawson:
fixes #1018

### DIFF
--- a/src/routes/_components/list/ListItem.html
+++ b/src/routes/_components/list/ListItem.html
@@ -5,13 +5,5 @@
               virtualIndex={index}
               virtualLength={length}
               virtualKey={key}
-              active={active}
   />
 </div>
-<script>
-  export default {
-    computed: {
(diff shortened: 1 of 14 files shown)


nolanlawson:
I decided to go with the background color change as the focused indicator, mostly for aesthetics. Also this makes it work almost exactly the same as it did before, except that the activeElement is correctly set.

I'm a bit on-the-fence about how to handle `j` when you already have the focus inside of a status (I don't like that it goes to the first element instead of the "next" element), but it's not terrible and it's at least consistent with what we had before.

PR #1033, opened 2019-02-23
fix: fix keyboard shortcuts for pinned toots

### LINKED ISSUE #908: j/k keyboard shortcuts don't work on pinned toots
1. Go to someone's profile who has pinned toots
2. Tap <kbd>j</kbd>
3. The first non-pinned toot is activated, not the first pinned toot

nolanlawson:
fixes #908

### DIFF
--- a/src/routes/_components/list/List.html
+++ b/src/routes/_components/list/List.html
@@ -10,7 +10,6 @@
     />
   {/each}
 </div>
-<ScrollListShortcuts bind:items=safeItems/>
 <style>
   .the-list {
     position: relative;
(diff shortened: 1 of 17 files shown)


nolanlawson:
Decided to largely rewrite ShortcutsScrollList, make it use a simple class name and attribute lookup in the DOM in order to work. Seems to be easier to manage state this way, especially since keyboard shortcuts can span multiple lists and types (pinned statuses, regular statuses, notifications).

nolanlawson:
Also I was not a fan of having shortcuts handled by the VirtualList; it doesn't seem like those two things should be related.

PR #1039, opened 2019-02-23
fix: fix timeline jumping on focus

### LINKED ISSUE #840: viewport jumps to previously-opened reply box when scrolling
1. open reply box for a toot
2. unfocus it
3. scroll up or down far enough to unload it from the virtual list
4. scroll back enough for it to be loaded again

the reply box will steal focus and the viewport will jump to it. sometimes the viewport immediately jumps back a frame later, sometimes not. not sure how to reproduce that part

i can reproduce this in chrome 71, firefox 64, and firefox for android 64

funny story, i had ran into this bug before but only looked into it now because i was backspacing in one reply box, the box got smaller by one line, which was enough to load another reply box lower down the page, which stole my keyboard focus while i was still backspacing 😓

nolanlawson:
This should fix #840. I can't see any reason why we would want it to scroll when we're focusing these things.

### DIFF
--- a/src/routes/_components/compose/ComposeInput.html
+++ b/src/routes/_components/compose/ComposeInput.html
@@ -88,7 +88,7 @@
             firstTime = false
             let { autoFocus } = this.get()
             if (autoFocus) {
-              requestAnimationFrame(() => textarea.focus())
+              requestAnimationFrame(() => textarea.focus({ preventScroll: true }))
             }
           }
         })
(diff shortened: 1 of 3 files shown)


nolanlawson:
Note that `focus({ preventScroll: true })` only seems to work in Chrome currently, but I can't think of a good way to fix this in Firefox/Safari. Hopefully they'll implement the spec soon. https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#Parameters

PR #1041, opened 2019-02-23
fix: use static server for `node server.js`

### LINKED ISSUE #1019: Unify node server and static server
`yarn build` should run the export command and update now.json.

`node server.js` should read from now.json and create a static server based on that.

This should eliminate discrepancies between the prod and self-hosted versions.

nolanlawson:
Aligns the Node server with the Zeit server. Fixes #1019

I admit this makes me a bit nervous as we're hacking around Sapper to do this, but it feels better to make the `node server.js` experience more-or-less exactly the same as the Zeit experience. Having two different apps here feels weird.


nolanlawson:
I'm nervous about this, but let's do it.

PR #1067, opened 2019-03-03
perf: use a separate icons.svg file

nolanlawson:
This splits icons into inline and non-inline. The inline ones are high
priority; the rest go in an icons.svg file.


nolanlawson:
This seems to mostly improve performance in terms of non-logged out users on the first page load. The HTML file is reduced from 46.3kB non-gzipped to 24.6kB, so it shaves off roughly 150ms on the transfer time on slow 3G.

It also reduces the total number of DOM nodes, although I don't see a decrease in memory usage.

It also reduces some repetition in the various HTML files which will mostly only impact the on-disk size for self-hosters, and even then not by much. (30522677 bytes total vs 30151196 bytes total, 371kb total)

I'm sort of mulling over this PR as it's a pretty marginal perf improvement and complicates the code quite a bit. So I'm trying to decide if it's worth it.

nolanlawson:
Now that I'm routing all SVGs into a single SvgIcon.html component, I feel a bit better about this. It's not so fragile anymore.

nolanlawson:
And now that I've fixed the TODO, I'm much happier with this

nolanlawson:
This also slightly reduces the size of all JS assets, probably due to moving all the SVG logic into one place. Before: 930202 After: 926318

PR #1069, opened 2019-03-03
perf: build url regex in advance

nolanlawson:
This actually slightly increases the size of measureText.js (from 11.56kB to 12.25kB) but I don't like the idea of having all this runtime logic to build up a static regex. Also the gzip size is smaller (6.85B to 6.33kB).


nolanlawson:
new size: 11.76kB, 6.3kB gzipped

PR #1070, opened 2019-03-03
fix: another attempted fix for thread dupliates

nolanlawson:


nolanlawson:
This code shoudn't be necessary; I'm not sure how to repro this bug, but this isn't the fix

PR #1074, opened 2019-03-04
fix: use correct sorting for snowflake IDs

### LINKED ISSUE #1071: Fix new ID format in Pleroma
Pleroma is moving to a system where IDs are no longer big ints represented as strings. This will surely break Pinafore because there are lots of places where it's assumed that we're dealing with big ints.

Related: https://github.com/tuskyapp/Tusky/pull/1016

nolanlawson:
fixes #1071

This may need IDB tests somehow, but based on manual testing it appears to be working.


nolanlawson:
I can do the IDB tests in another branch. It actually seems like a fair amount of work, now that I'm looking at it. But this seems solid so far anyway.

PR #1078, opened 2019-03-05
fix: add a privacy policy

### LINKED ISSUE #1077: Add a privacy policy
It would be good to clarify that Pinafore is a static site and all data is stored client side.

nolanlawson:
fixes #1077


nolanlawson:
FWIW I checked out the Now logs and confirmed that there's nothing in there except some file stats when uploading the files

PR #1083, opened 2019-03-06
Fix reversed base62 encoding

uiri:
Pleroma uses numbers, upper case, and then lower case for its ordering for base62.


nolanlawson:
Awesome thank you, I just puzzled through this myself in #1084. I'll merge yours and then merge mine as well because mine has an extra IDB migration to fix this. Thanks!

PR #1089, opened 2019-03-08
Fix author mention in posts that aren't replies

uiri:
It appears that, at least Pleroma, always includes the author of a post in the list of mentions. This appears to cause two, related issues:

1. In a post with a CW that is not a reply, the author will be shown in StatusMentions even though the post is not a reply and even if the status content does not contain an explicit self-mention.
2. When clicking reply, the compose dialog auto-populates two mentions of the author of the in-reply-to post. It makes sense to ensure that the author goes into the mentioned accounts, so it makes sense to filter out the author if they are also listed in the mentions array.

### DIFF
--- a/src/routes/_actions/compose.js
+++ b/src/routes/_actions/compose.js
@@ -11,7 +11,7 @@ export async function insertHandleForReply (statusId) {
   let status = await database.getStatus(currentInstance, statusId)
   let { currentVerifyCredentials } = store.get()
   let originalStatus = status.reblog || status
-  let accounts = [originalStatus.account].concat(originalStatus.mentions || [])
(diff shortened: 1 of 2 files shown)


nolanlawson:
I've noticed this bug and didn't know what was causing it. Thanks for identifying!

It looks, though, like this PR breaks one of our tests ([see test log here](https://travis-ci.org/nolanlawson/pinafore/jobs/503432597)), which is the one that populates an @-mention when replying to an account. So maybe there is a fundamental incompatibility between Mastodon and Pleroma here? Is there some solution that will work for both Mastodon and Pleroma?

nolanlawson:
This seems fixed in recent versions of Pleroma. Please reopen if I'm incorrect.

Related: https://github.com/nolanlawson/pinafore/pull/1592

PR #1094, opened 2019-03-10
perf: switch from inferno to preact, use emoji-mart prerelease

nolanlawson:
This switches us to a prerelease version of emoji-mart, using a branch with many of my open PRs https://github.com/missive/emoji-mart/pulls

It also switches from inferno to preact, reducing the total size of all JS files from 898.41kb to 882.99kb, mostly due to preact being smaller then inferno (18.88kb vs 35.46kb).

I'm mostly just opening this PR to see if the tests all pass.


nolanlawson:
Sweet, looks like preact@10 fixes a compat issue where the search text was being prepopulated (and thus breaking the test).

Related: https://github.com/missive/emoji-mart/pull/290

nolanlawson:
As an added bonus, the total JS size is down to 877.87kb because preact is even smaller than before (14.25kb)

nolanlawson:
Doesn't seem to be any harm in trying this out on dev.pinafore.social, looks good so far based on my testing

PR #1098, opened 2019-03-16
perf: remove stats.json from webpack

nolanlawson:
I was not ever really looking at this, and it seems to add compilation time


nolanlawson:
Yeah this cuts the `sapper build` time from ~30s to ~15s.

PR #1119, opened 2019-03-26
fix: remove no longer needed `initial-scale=1`

alrra:
`initial-scale=1` was a workaround for a orientation change bug, that was only needed for the no longer relevant Safari for iOS < 9.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

Ref:

  * https://www.quirksmode.org/blog/archives/2013/10/more_about_scal.html
  * https://twitter.com/ppk/status/829329567219343360

### DIFF
--- a/src/build/template.html
+++ b/src/build/template.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset='utf-8' >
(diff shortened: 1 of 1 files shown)


nolanlawson:
Thanks!

PR #1120, opened 2019-03-27
Enable PWA support for iOS devices

### LINKED ISSUE #1058: Test PWA support in iOS 12.2
Apparently we may be able to remove the manifest.json removal in iOS 12.2. They have a new PWA implementation.

Background: https://github.com/nolanlawson/pinafore/issues/45 and https://toot.cafe/@tomayac/101671479465139470

sgenoud:
With the release of iOS 12.2, the OAuth PWA workflow can work. I tested it on my iPhone and iPad.

Closes #1058

### DIFF
--- a/src/inline-script/inline-script.js
+++ b/src/inline-script/inline-script.js
@@ -50,3 +50,13 @@ if (hasLocalStorage && localStorage.store_disableCustomScrollbars === 'true') {
 if (/mac/i.test(navigator.platform)) {
   document.documentElement.style.setProperty('--scrollbar-border-radius', '50px')
 }
+
+// Versions of iOS Safari before iOS 12.2 do not work properly as a PWA
+// for cross-origin authentication: https://github.com/nolanlawson/pinafore/issues/45
+// Here we sniff for iOS <12.2 by checking for the existence of a native IntersectionObserver
+// function, which was added in 12.2.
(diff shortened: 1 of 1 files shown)


nolanlawson:
I'd be more comfortable if we did a UA sniff to only do this for iOS 12.2 and above. 12.2 is super recent and so not every iOS device will have it immediately.

sgenoud:
What would you think of just enabling it for dev.pinafore for a little while. It would let us figure out if everything is usable as a pwa (if you are brave enough), while not breaking anything for normal users. 

Adoption of new versions is quite good on iOS, in a month or two we can probably assume that most people have upgraded.

nolanlawson:
I'd be happy to just merge this pull request and push it to dev.pinafore.social, if we added the UA sniff. :)

I know iOS users upgrade quickly, but since it's only a line of code or so to do the UA sniff, I'd still prefer to do that.

12% of iOS users are still on iOS 11: https://developer.apple.com/support/app-store/

nolanlawson:
^ Here's my suggestion: we sniff for a native IntersectionObserver implementation to test for iOS 12.2.

nolanlawson:
OK, let's try this out. It will be live on dev.pinafore.social soon.

sgenoud:
Thanks! I was not up for a regex challenge but the intersection observer was a great idea!

PR #1121, opened 2019-03-27
Add support for Open Graph cards

### LINKED ISSUE #740: Opengraph cards
Copied from #6 

I'm kind of on-the-fence about this feature, but it's worth investigating. I guess it would involve fetching the URL and parsing it for the opengraph embedded image. For some servers it may not be possible due to CORS.

sgenoud:
Closes #740. The card info is returned from the Mastodon interface, so it is easier than what was argued there.

Capture d'écran 2019-03-27 19 16 41

Capture d'écran 2019-03-27 19 16 09

### DIFF
--- a/src/routes/_components/status/StatusCard.html
+++ b/src/routes/_components/status/StatusCard.html
@@ -1,0 +1,91 @@
+<a href="{url}" class="status-card" target="_blank" rel="noopener noreferrer">
+    <span class="card-title">
+     {title}
+    </span>
+  {#if hasBody}
+    <div class="card-content">
+  {#if imageUrl}
+    <LazyImage forceSize=true height="50" width="50" src={imageUrl} />
+  {/if}
+  {#if description}
+    <span class="card-description">
+      {description}
+    </span>
+  {/if}
+  </div>
+  {/if}
+</a>
+<style>
+  .status-card {
+    grid-area: card;
+    display: flex;
+    flex-direction: column;
+    padding: 15px;
+
+    text-decoration: none;
+    color: inherit;
+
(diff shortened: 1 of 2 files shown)


src/routes/_components/status/StatusCard.html
```diff
@@ -0,0 +1,91 @@
+<a href="{url}" class="status-card" target="_blank" rel="noopener noreferrer">
+    <span class="card-title">
+     {title}
+    </span>
+  {#if hasBody}
+    <div class="card-content">
+  {#if imageUrl}
+    <LazyImage forceSize=true height="50" width="50" src={imageUrl} />
```
nolanlawson: Is there an `alt` we can pass in to this image?

nolanlawson: Alternatively, we can make it aria-hidden if it's just decorative. I don't know much about open graph.

nolanlawson: Yeah OK, I just checked a few, and it seems there is no alt. There is a `description`, but that seems more like the meta description tag. So we can just make it aria-hidden.

sgenoud:
Thanks for the change and the merge. I looked into the alt text for open graph, it seems that the it supports it but mastodon does not fetch the information. 

Do you have a reference for a good (and pragmatic) intro on the most important aria arguments, I have never spent the time learning this part of the web yet.

nolanlawson:
I've mostly learned this stuff by playing around with VoiceOver (on macOS) and reading https://www.w3.org/TR/WCAG21/ :)

PR #1131, opened 2019-04-02
fix: add placeholder icon for opengraph cards without images

nolanlawson:
I don't like that opengraph cards without images don't show any description. So this adds a little placeholder icon like what Mastodon uses.


nolanlawson:
Screenshot from 2019-04-01 19-59-28

PR #1132, opened 2019-04-02
Add a setting to reload the page

sgenoud:
The iOS PWA version does not allow you to reload the page - even with killing the app. 

With this feature the user can reload for new versions of the app. It makes it possible to solve bad transient state issues as well.

### DIFF
--- a/src/routes/_pages/settings/index.html
+++ b/src/routes/_pages/settings/index.html
@@ -12,7 +12,7 @@
       <SettingsListButton href="/settings/hotkeys" label="Hotkeys"/>
     </SettingsListRow>
     <SettingsListRow>
-    <SettingsListButton href="" on:click=reload(event) label="Reload"/>
+    <SettingsListButton href="#" on:click="reload(event)" label="Reload"/>
     </SettingsListRow>
     <SettingsListRow>
       <SettingsListButton href="/settings/about" label="About Pinafore"/>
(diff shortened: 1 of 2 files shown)


nolanlawson:
Thanks for the PR! Although this is kind of an odd place to put the reload button. I think I would prefer:

- instead of a toast, the "update available" message is a permanent alert that slides up at the bottom of the page and is fixed, and it has a "reload" button inside of it
- for the transient state issues, maybe a "pull to refresh" or a refresh button at the top of the timeline. I know there are bugs here, but I haven't been able to nail them down yet.

sgenoud:
I agree that the message in the toast would be a better solution - for asking people to reload. And a pull to refresh would be nice.

Both are heavier in terms of implementation - my aim was to try to make the app quickly usable as a PWA on iOS.

Would you agree on hiding this feature somewhere else, while we work on a better solution for the long term?

Note that the transient issues are worse on iOS PWA as well. I guess that there is a problem (propably on Apple's side) on setting back the state of the websockets after the app goes into some kind of sleep.

src/routes/_pages/settings/index.html
```diff
@@ -11,6 +11,9 @@ <h1>Settings</h1>
     <SettingsListRow>
       <SettingsListButton href="/settings/hotkeys" label="Hotkeys"/>
     </SettingsListRow>
+    <SettingsListRow>
+    <SettingsListButton href="" on:click=reload(event) label="Reload"/>
```
nolanlawson: This href should be a `#` if it's a dynamic event. I forget why, but I think this is what you're supposed to do :joy:

nolanlawson:
Hmmm, I think you're right. This is a decent workaround, and there are lots of cases where a PWA implementation doesn't make it easy to reload the page (not just iOS), so it's useful in more than one case. Also it's fairly hidden in the Settings as-is. OK, let's do it. :+1:

PR #1133, opened 2019-04-02
Disable custom scrollbars by default on iOS

### LINKED ISSUE #1085: Scrollbar flash in iOS Safari
We may just want to sniff iOS and disable the custom scrollbar in that case: 

> If, like me the scrollbar flash in @pinafore on your iOS device has been bugging you for a long time, do not blame pinafore, blame Safari. 
> 
> It seems like there is a bug in how it manages styled scrollbars. Just disable custom scrollbars (in Settings) and it will be solved.

sgenoud:
Fixes #1085


nolanlawson:
Makes sense to me! Thanks

PR #1134, opened 2019-04-02
Add a shortcut to opening the cards link

sgenoud:


src/routes/_components/status/StatusCard.html
```diff
@@ -90,6 +95,12 @@
       url: ({ card }) => card.url,
       description: ({ card }) => card.description || card.provider_name,
       imageUrl: ({ card }) => card.image
+    },
+    methods: {
+      open (url) {
+        window.open(url)
```
nolanlawson: Using `window.open()` in this way is a [security vulnerability](https://mathiasbynens.github.io/rel-noopener/). We should do a `.click()` on the `<a>` element instead, which AFAIK is [the only reasonable way to get the rel=noopener behavior we want](https://stackoverflow.com/a/47644052).

sgenoud: Thansk, I always added the noopener without understanding why I was adding it. This is now fixed!

nolanlawson:
fixed merge conflicts, waiting for green test run

PR #1135, opened 2019-04-02
Remove placeholder for cards without images

sgenoud:
It feels like a waste of space, especially when shown on mobile devices (as in the picture).

Capture d'écran 2019-04-02 09 34 11
Capture d'écran 2019-04-02 09 34 40


nolanlawson:
You're right, this is one case where it probably doesn't make sense to follow design of Twitter/Mastodon. I guess it makes it visually consistent, but it is definitely a waste of space.

We can also remove the SVG entirely from `svgs.js` now that it's no longer used.

Issue #785, opened 2018-12-12
[Feature request] Option to send direct messages via the hamburger menu


Hund:
Currently there's 6 steps to send a direct message when you're on someones profile:

`Menu / Mention @<user> / Change privacy setting to "direct" / Type message / Send message`

But if there's a option via the hamburger menu to directly send a message (like it's possible via their regular official web interface) it would only require 4 steps. And it would also make it a bit safer, I don't send any *private* messages (in the context of private information), but I sometimes forget to change the privacy settings and I post messages only meant for two sets of eyes. :)

Would this be something you would consider adding?

nolanlawson:
So there is currently a "mention so-and-so" option in the profile menu dropdown. I.e. the steps are: click on profile, click on hamburger, click on "mention," click on globe icon, click on "direct." Maybe this is too undiscoverable, though?

Alternatively, maybe this could go in the dedicated DM UI. (https://github.com/nolanlawson/pinafore/issues/639) What do you think?

Hund:
That's what I was talking about. I don't think it's undiscoverable, it just requires two more steps than needed. :)

This is how it looks on Mastodons official UI:

![directmessage](https://user-images.githubusercontent.com/320498/49925909-8c899880-feba-11e8-84a3-808d2969c333.png)

strypey:
I just want to see this bet, and raise it. A fediverse newbie who's just figured out how to follow someone has already gone through quite a few steps, and could be a bit flummoxed. I'd like to see explicit buttons on profile pages for 'Mention User' and 'Send Direct Message'. Maybe the UI could have a line for this beneath the 'Toots | Follows | Followers' line, and the hamburger menu button could be on that line too? I think that would help it stand out more.

Extra for experts. As a user I'd love to have a user setting that allowed DMs to be sent to the email address associated with my fedi account, instead of (or as well as) appearing as a notification in Pinafore. I've filed a separate feature request on this (https://github.com/nolanlawson/pinafore/issues/1615) but I mention it here as its kind of related.

achou11:
adding yet another opinion here because after using Pinafore for a reasonably long period of time, I only just learned how to create a DM 😅 not fixated on the proposed fix but i do wish it were easier to figure out how to do this

alternatively, update the user guide to show how that's done in its current state

PR #1137, opened 2019-04-02
Adds a basic conversations timeline

### LINKED ISSUE #639: Separate UI for DMs
This is something supported in Mastodon 2.6.1 now but we need to test it out on multiple Mastodon versions. As of Mastodon 2.6.1, DMs are no longer in the Home timeline, so this is probably a must-have now since otherwise they're hard to find (only in notifications).

sgenoud:
Capture d'écran 2019-04-02 12 09 35

Closes #639 (but not #785)

The support is not perfect, I took a couple of shortcuts in order to avoid having to reproduce what you had to do with notifications. Conversations should be their own objects with everything that goes with it. As a first implementation, I would go with something simple, and we can iterate if we need a better DM UI.

### DIFF
--- a/src/routes/_store/computations/navComputations.js
+++ b/src/routes/_store/computations/navComputations.js
@@ -24,12 +24,12 @@ export function navComputations (store) {
           svg: '#fa-globe',
           label: 'Federated'
         }
-      } else if (pinnedPage === '/conversations') {
+      } else if (pinnedPage === '/direct') {
         pinnedPageObject = {
-          name: 'conversations',
-          href: '/conversations',
+          name: 'direct',
+          href: '/direct',
           svg: '#fa-envelope',
-          label: 'Conversations'
+          label: 'Direct messages'
         }
       } else if (pinnedPage === '/favorites') {
         pinnedPageObject = {
--- a/src/routes/_store/computations/navComputations.js
+++ b/src/routes/_store/computations/navComputations.js
@@ -24,6 +24,13 @@ export function navComputations (store) {
           svg: '#fa-globe',
           label: 'Federated'
         }
+      } else if (pinnedPage === '/conversations') {
+        pinnedPageObject = {
+          name: 'conversations',
+          href: '/conversations',
+          svg: '#fa-envelope',
+          label: 'Conversations'
+        }
       } else if (pinnedPage === '/favorites') {
         pinnedPageObject = {
           name: 'favorites',
--- a/src/routes/conversations.html
+++ b/src/routes/conversations.html
@@ -1,0 +1,20 @@
+<Title name="Favorites" />
+
+  <LazyPage {pageComponent} {params} />
+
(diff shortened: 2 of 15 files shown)


nolanlawson:
Thanks! I will be on vacation for awhile, but I will look at this soon. I agree doing the quickest possible thing is probably best. There are lots of cases where it's awkward to fit something into the "timeline" model (e.g. favorites, threads), but so far it's worked okay.

sgenoud:
Enjoy your holidays!

src/routes/_components/compose/ComposeBox.html
```diff
@@ -43,10 +43,6 @@ <h1 class="sr-only">Compose status</h1>
     max-width: calc(100vw - 40px);
   }
 
-  .compose-box.direct-reply {
-    background-color: var(--status-direct-background);
-  }
-
```
nolanlawson: I think if we remove this, then it's not very obvious which messages are DMs and which aren't, when we're not in the "conversations" timeline.

I think maybe this color could be set for DMs only when we're _not_ in the conversations timeline.

src/routes/conversations.html
```diff
@@ -0,0 +1,20 @@
+<Title name="Favorites" />
```
nolanlawson: This name shouldn't be "Favorites"

src/routes/_store/computations/navComputations.js
```diff
@@ -24,6 +24,13 @@ export function navComputations (store) {
           svg: '#fa-globe',
           label: 'Federated'
         }
+      } else if (pinnedPage === '/conversations') {
+        pinnedPageObject = {
+          name: 'conversations',
+          href: '/conversations',
+          svg: '#fa-envelope',
+          label: 'Conversations'
+        }
```
nolanlawson: In general, I prefer to match Mastodon's UI and URL conventions whenever possible. It seems that the Mastodon web interface uses `/direct` for the URL and "Direct messages" for the user-facing text. So I'd prefer to stick to that if possible.

nolanlawson:
Fixed merge conflicts

nolanlawson:
I had some free time so I made the suggested changes. I'll merge it as two commits to preserve authorship. Thanks a bunch!

nolanlawson:
Merged as 622dbde258dcb01ba1e7fe5879880898c18ff193 and b3098c8c5a875cbecc1d5d8d3a6fb4489f49402d

sgenoud:
> I think maybe this color could be set for DMs only when we're not in the conversations timeline.

yeah, good point. With your changes if you look at the thread of a conversation if all has the color (it does not feel right). I might make a PR at some point.

I hesitated to call them direct messages as well. There was a bit of terminology in Mastodon that call it conversations. I prefered the way that sounded. Good with DM as well, though.

Thanks for the changes!

PR #1138, opened 2019-04-04
Manage the width to take scrollbars into account

sgenoud:
There was a small visual bug when trying to have a small column in an environment with scrollbars (for instance small desktop PWA).

Before:

Before

After:

After

### DIFF
--- a/src/routes/_components/dialog/components/MediaDialog.html
+++ b/src/routes/_components/dialog/components/MediaDialog.html
@@ -79,7 +79,7 @@
 <Shortcut scope='modal-{id}' key="ArrowRight" on:pressed="next()" />
 <style>
   :global(.media-modal-dialog) {
-    max-width: calc(100vw);
+    max-width: calc(100%);
   }
   .media-container {
     height: calc(100% - 64px); /* 44px X button height + 20px padding */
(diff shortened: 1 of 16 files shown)


nolanlawson:
Makes sense to me; my design in this area has always been kind of wonky. I'll give this a test drive soon, but LGTM.

nolanlawson:
Unfortunately this doesn't work well for the compose dialog on mobile Chrome (using Chrome's responsive design simulation):

Screenshot from 2019-04-08 19-18-30

I have a special set of styles for this mode (called `slim-size`) which the new design doesn't seem to take into consideration.

src/routes/_components/dialog/components/MediaDialog.html
```diff
@@ -79,7 +79,7 @@
 <Shortcut scope='modal-{id}' key="ArrowRight" on:pressed="next()" />
 <style>
   :global(.media-modal-dialog) {
-    max-width: calc(100vw);
+    max-width: calc(100%);
```
nolanlawson: I'm not sure why I had a `calc()` here

nolanlawson:
Fantastic work. I guess I don't need the "slim" sizing anymore. I gave this a try and everything looks great. Thanks a bunch!

PR #1145, opened 2019-04-09
Add hostname when description is missing

### LINKED ISSUE #1136: Card look and feel for different information
As related to #1135, as well as a private discussion let's discuss what are the different possibilities of information present:

- [x] title, image and description (or provider name) (case a)
- [x] title, description (or provider name) –  image missing (case b) #1135 
- [ ] title, image – description (or provider name) missing (case c)
- [ ] title only – description (or provider name) and image missing (case d)

Do we agree on the case b?

For the case c, I would propose two different solutions. Either put the domain if there is not provider name (so there is always a description text) or just hide the image.

Personally I do not mind the title only case if the data is empty, but there might be other opinions?

sgenoud:
Closes  #1136

Capture d'écran 2019-04-09 16 51 42

### DIFF
--- a/src/routes/_components/status/StatusCard.html
+++ b/src/routes/_components/status/StatusCard.html
@@ -81,7 +81,8 @@
       card: ({ originalStatus }) => originalStatus.card,
       title: ({ card }) => card.title,
       url: ({ card }) => card.url,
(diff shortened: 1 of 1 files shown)


nolanlawson:
Thanks!

PR #1146, opened 2019-04-09
First attempt at a pull to refresh mechanism (experimental)

sgenoud:
I used [this library](https://github.com/BoxFactura/pulltorefresh.js) that seems well maintained and claims to be lightweight. It actually made the implementation fairly easy. Note that I have never used it before - so there might be stuff hiding in the shadows.

For now I added a pull to refresh on the main timeline - it restarts the stream and fetches the latest element (not sure if it is a good idea).

I could see adding it to other pages. For instance on the settings to trigger the reload (it would be a bit more hidden that way).

For reference, this is a partial follow up to #1132

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "pinch-zoom-element": "^1.1.1",
     "preact": "^10.0.0-alpha.2",
     "prop-types": "^15.7.2",
-    "pulltorefreshjs": "^0.1.16",
+    "pulltorefreshjs": "^0.1.17",
     "quick-lru": "^4.0.0",
     "remount": "^0.9.5",
     "requestidlecallback": "^0.3.0",
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "pinch-zoom-element": "^1.1.1",
     "preact": "^10.0.0-alpha.2",
     "prop-types": "^15.7.2",
+    "pulltorefreshjs": "^0.1.16",
     "quick-lru": "^4.0.0",
     "remount": "^0.9.5",
     "requestidlecallback": "^0.3.0",
--- a/src/routes/_actions/timeline.js
+++ b/src/routes/_actions/timeline.js
@@ -80,7 +80,7 @@ export async function addTimelineItemSummaries (instanceName, timelineName, newS
   }
 }
 
-async function fetchTimelineItemsAndPossiblyFallBack () {
+export async function fetchTimelineItemsAndPossiblyFallBack () {
   mark('fetchTimelineItemsAndPossiblyFallBack')
   let {
     currentTimeline,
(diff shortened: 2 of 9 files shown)


nolanlawson:
Thanks for opening this PR! I have definitely been meaning to start looking into pull-to-refresh, especially given the transient state issues (e.g. when you haven't opened the app for awhile, then you open it, and for whatever reason the web socket is no longer listening).

One main issue I have is that I would like to implement pull-to-refresh in such a way that it doesn't de-optimize scrolling. If you open up Chrome Dev Tools -> Rendering -> Scrolling performance issues, you'll see that this library adds touchmove/touchstart listeners to the `window`:

https://github.com/BoxFactura/pulltorefresh.js/blob/b8ce09503f9b121db280cff8faa1ae790809a90e/src/lib/events.js#L133-L134

This de-optimizes scrolling, because now all touch events have to go through the main thread. So the browser is not able to move scrolling to a background thread nearly as efficiently.

I was really hoping there was a library out there somewhere that used Pointer Events instead of Touch Events, but maybe it doesn't exist?

sgenoud:
I forked the library, and made a PR: https://github.com/BoxFactura/pulltorefresh.js/pull/70 Hope they are interested in the changes. 

Thanks for telling me about how the pointer events are better!

sgenoud:
I forked the library, and made a PR: https://github.com/BoxFactura/pulltorefresh.js/pull/70 Hope they are interested in the changes. 

Thanks for telling me about how the pointer events are better!

nolanlawson:
Excellent work on getting pointer events into that library. Looks like they've merged and published it.

The more I think about it, the more I think you're right that we need some kind of mechanism like this. Unfortunately even with my page lifecycle fixes which activate on freeze/unfreeze events, there are still lots of situations where you open the browser after a long time away, and the results are not fresh. (E.g. when you stop streaming and then start again... how do you fill in that "gap"? It's a non-trivial problem.)

I'm still concerned about accessibility and performance, but I'm going to keep looking into this. Maybe we can add an explicit button as an option for screenreader users or keyboard users.

nolanlawson:
When I play around with this PR using a mousewheel in Chrome, it seems to get "stuck" occasionally, where if you pull-to-refresh, then click on another nav button, the pull-to-refresh message is already there. Perhaps we should only enable this for touchscreen users or something... it needs some testing.

nolanlawson:
The "stickiness" repros in desktop Firefox with a mousewheel as well. Also it would be nice if the message appeared at the top of the timeline, whereas in the home timeline it appears above the compose box, and in the notification timeline it appears above the "all/mentions" tab...

sgenoud:
The base issue was fixed with #1349

PR #1154, opened 2019-04-14
fix: fix dialogs within dialogs not opening

### LINKED ISSUE #1153: Can't open mute dialog twice
1. Click the ... on a status
2. Click "mute"
3. Press Esc to close the MuteDialog
4. Repeat steps 1-2
5. The MuteDialog doesn't show up, even though it should.

I imagine this is caused by #826

nolanlawson:
fixes #1153

### DIFF
--- a/src/routes/_components/dialog/components/GenericConfirmationDialog.html
+++ b/src/routes/_components/dialog/components/GenericConfirmationDialog.html
@@ -43,7 +43,8 @@
     },
     data: () => ({
       positiveText: void 0,
-      negativeText: void 0
+      negativeText: void 0,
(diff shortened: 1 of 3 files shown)


src/routes/_components/dialog/components/GenericConfirmationDialog.html
```diff
@@ -43,7 +43,8 @@
     },
     data: () => ({
       positiveText: void 0,
-      negativeText: void 0
+      negativeText: void 0,
+      className: ''
```
nolanlawson: this is just to avoid a dev console warning

PR #1168, opened 2019-05-02
perf: render virtual list items lazily

nolanlawson:
Based on my testing on Chrome on Android, this is a modest performance improvement. Rather than greedily render all virtual list items, we render them one-at-a-time inside of requestIdleCallback, being careful to render them in the right order so that it doesn't look weird.

This speeds up initial rendering of the timeline, because you don't have to wait for 20 items to render; you only have to wait for the top 1-3 or so.

### DIFF
--- a/src/routes/_components/virtualList/VirtualListLazyItem.html
+++ b/src/routes/_components/virtualList/VirtualListLazyItem.html
@@ -9,15 +9,20 @@
 <script>
   import VirtualListItem from './VirtualListItem'
   import { mark, stop } from '../../_utils/marks'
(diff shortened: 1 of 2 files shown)


nolanlawson:
Unfortunately it looks like the tests are failing due to focus resolution not working (e.g. after closing a modal, after pressing the back button). I can't think of a good way to solve both problems. We would need to delay the focus() call until after requestIdleCallback, and probably try it a few times before bailing out. Focus management makes this really difficult to implement.

nolanlawson:
Gonna reopen

PR #1172, opened 2019-05-03
perf: render virtual list items lazily

nolanlawson:
reopening #1168 because I figured out the focus issues


nolanlawson:
This feels somewhat like a mixed bag, because by breaking things up into smaller chunks of work, we end up doing more work overall. But it does seem to result in faster rendering of the items at the top of the timeline, anyway. I worry it's an overoptimization that will complicate things in the long run.

nolanlawson:
Yeah, this is a cute idea, but I think I'm going to punt on it. It makes initial load slower overall, even if it makes subsequent loads less janky. There is no magic bullet here - only hard-to-quantify tradeoffs.

PR #1181, opened 2019-05-05
chore: update to mastodon v2.8.2

nolanlawson:


nolanlawson:
Turns out the test was wrong; when a locked account boosts you, it shouldn't increment the boost counter. Mastodon fixed this bug, so the test started failing.

PR #1190, opened 2019-05-06
fix: throttle XHRs from autosuggest

nolanlawson:
We should be a good citizen and not spam the server with one XHR per keystroke


nolanlawson:
On second thought we should also cancel inflight XHRs as well. The Mastodon frontend does this.

PR #1197, opened 2019-05-09
fix: fix style of snackbar on mobile

nolanlawson:


sgenoud:
Thanks, I just saw this was an issue on my latest iPhone particularly as the corner are cropped (in addition to the text being very close on the right. Could you add some additional padding on the bottom as well?

IMG_0708

(my eye considers the white bar as the end of the screen somehow)

nolanlawson:
Interesting, yeah the bottom could use more padding too

nolanlawson:
@sgenoud Is that white bar specific to the iPhone X? Is this a good description of the issue? https://css-tricks.com/the-notch-and-css/

nolanlawson:
Merging for now, I'd like to address the iPhone issue separately since there may be a way to detect it in CSS

sgenoud:
> Is that white bar specific to the iPhone X? 

To the X, and the full latest generation (Xr, Xs). So I guess all iPhones in the near future.

> Is this a good description of the issue? https://css-tricks.com/the-notch-and-css/

The article you sent is more related to the notch whitch is at the top of the screen - this is only an issue in landscape mode (and I just realised it now, I never use pinafore in landscape mode).
At the bottom of the screen you also have an "unsafe area" that is needed to be taken care of.

Nevertheless the article gives the tools to solve the issue. You probably just need to add the `safe-area-inset-bottom` in the bottom padding of the snackbar.

For the top of the screen, this is still a nightmare – I actually tried to fix it a little while ago for pinafore and decided that the tools to have something better are not there yet, and to wait for improvements on Apple's side.

PR #1199, opened 2019-05-09
Fix ios design issues

### LINKED ISSUE #1198: Fix snackbar style relative to iPhone notch
As described in https://github.com/nolanlawson/pinafore/pull/1197#issuecomment-490758966, the iPhone notch can appear to cut off the snackbar a bit.

sgenoud:
Change the viewport behaviour to work with the iPhone X like screens (closes #1198).

While I was in improving design for iOS, I also changed the default icons - the old one where missing some white space. Feel free to revert. The icon on the right is the new one.

IMG_62CFAE227B23-1


nolanlawson:
Awesome, thank you!! This looks great. I will defer to your expertise on this, because I don't have an iPhone X; I only have an iPod Touch (6th gen) for testing. :) I agree the new icon looks much better with more whitespace.

Issue #14, opened 2018-04-02
Sometimes new statuses/notifications don't show up


nolanlawson:
Sometimes new statuses and notifications don't appear until you refresh. Not sure how this happens, but I think it may be related to the browser suspending background tabs.

spriteclad:
Thought I might as well throw what happened to me in the ring -- turns out my adblocker (uBlock Origin's EasyPrivacy list) was blocking all WebSocket requests from any external domain ending in `.space`. Yeeeaaah. Might want to either disable your ad blocker if you're also using Awoo and EasyPrivacy.

nolanlawson:
Based on testing in Firefox and Chromium, this appears fixed at least insofar as when you come back to the app after a long period of time the notifications appear, and you don't need to refresh. So probably either #1196 or #1195 (or both) fixed it.

nolanlawson:
Still not fixed, Chromium on Android

nolanlawson:
Still not fixed, according to a Chrome on macOS user. "I've gotten used to refreshing to make sure I see my notifications."

nolanlawson:
This is *still* happening. I just repro'd it on my wife's machine in Chrome.

What happened was this:

1. During the "active" state transition in lifecycle.js, the `this._ws` (websocket connection) was defined, and its `readyState` was [`WebSocket.OPEN` (1)](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket). So we didn't bother trying to reconnect. However, the websocket connection was not actually open... >_<

githubetc:
Just confirming that same "no auto refresh" bug is still happening with Pinafore version 1.14.0 on Chrome Android and Chrome MacOS.

Notifications are OK, got them almost instantly, it is just that the statuses do NOT show up on the timeline automatically. So I have to keep clicking refresh after hearing each the notification.

nolanlawson:
Can't consistently repro this, and it may be a bug in Mastodon itself. If I can get a consistent repro, then I'll look into it.

PR #1205, opened 2019-05-11
fix: use fibonacci instead of exponential when streaming

nolanlawson:
I dunno if this will improve #14, but it at least aligns us with Mastodon's behavior.

### DIFF
--- a/src/routes/_api/TimelineStream.js
+++ b/src/routes/_api/TimelineStream.js
@@ -71,7 +71,7 @@ export class TimelineStream {
 
   _setupWebSocket () {
     const url = getUrl(this._streamingApi, this._accessToken, this._timeline)
-    const ws = new WebSocketClient(url, null, { backoff: 'exponential' })
+    const ws = new WebSocketClient(url, null, { backoff: 'fibonacci' })
 
     ws.onopen = this._opts.onOpen || noop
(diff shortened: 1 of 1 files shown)


nolanlawson:
Well if nothing else, the fibonacci algorithm doesn't seem to increase as quickly as the exponential one (removing Math.randoms):

```
attempt 1 fib 100 exp 200
attempt 2 fib 200 exp 400
attempt 3 fib 300 exp 800
attempt 4 fib 500 exp 1600
attempt 5 fib 800 exp 3200
attempt 6 fib 1300 exp 6400
attempt 7 fib 2100 exp 12800
attempt 8 fib 3400 exp 25600
attempt 9 fib 5500 exp 51200
attempt 10 fib 8900 exp 102400
attempt 11 fib 14400 exp 204800
attempt 12 fib 23300 exp 409600
attempt 13 fib 37700 exp 819200
attempt 14 fib 61000 exp 1638400
attempt 15 fib 98700 exp 3276800
attempt 16 fib 159700 exp 6553600
attempt 17 fib 258400 exp 13107200
attempt 18 fib 418100 exp 26214400
attempt 19 fib 676500 exp 52428800
attempt 20 fib 1094600 exp 104857600
```

So indeed maybe this would cause problems reconnecting; you could end up waiting longer for a websocket to connect in Pinafore than in Mastodon.

PR #1211, opened 2019-05-11
feat: add a config.json file

nolanlawson:
This is a pretty straightforward change that I think makes admin'ing Pinafore a bit more convenient.

One thing I'm unsure of is how this would work with Docker. I have no Docker experience.


nolanlawson:
this would fix #1211

nolanlawson:
I don't want to add this if it's not going to help with the "locked instance" issue. However I do like splitting out "default" theme from "no external CSS required" theme.

PR #1213, opened 2019-05-12
fix: use only one 180x180px touch icon

alrra:
* Include just one 180x180px touch icon`.

  Over time as Apple released different size displays for their devices, [requirements](https://github.com/h5bp/html5-boilerplate/pull/1599#issuecomment-56384135) for the size of the touch icon have changed quite a bit:

   * 57×57px – iPhone with @1x display and iPod Touch
   * 72×72px – iPad and iPad mini with @1x display running iOS ≤ 6
   * 76×76px – iPad and iPad mini with @1x display running iOS ≥ 7
   * 114×114px – iPhone with @2x display running iOS ≤ 6
   * 120×120px – iPhone with @2x and @3x display running iOS ≥ 7
   * 144×144px – iPad and iPad mini with @2x display running iOS ≤ 6
   * 152×152px – iPad and iPad mini with @2x display running iOS 7
   * 180×180px – iPad and iPad mini with @2x display running iOS 8+

  However, [most iOS users will be on the latest 2 versions](https://developer.apple.com/support/app-store/) of iOS and using newer devices, so nowadays, one 180x180px touch icon is enough.

  Also, if needed, the icon [will be automatically downscaled](https://realfavicongenerator.net/blog/how-ios-scales-the-apple-touch-icon/) by Safari, and the result of the scaling is generally ok.

* Remove unneeded `sizes` attribute.

  When using only one touch icon there is no need to use the `sizes` attribute.

----

See also: https://mathiasbynens.be/notes/touch-icons


nolanlawson:
Cool, thanks! :tada:

Issue #1130, opened 2019-04-01
Implement polls


nolanlawson:
This is a new thing in Mastodon 2.8.

nolanlawson:
Docs: https://source.joinmastodon.org/mastodon/docs/blob/master/content/en/api/rest/polls.md

nolanlawson:
Sub-tasks:

- [x] ability to view poll results (fixed in #1219)
- [ ] ability to create a poll
- [ ] ability to interact with a poll
- [ ] push notification support for poll results

nolanlawson:
Poll notifications are still a bit busted. Looks like it should say something like "A poll you have voted in has ended." It says this even if you created the poll and didn't vote, but we can change that text.

nolanlawson:
Unless I'm missing something, push notifications for polls just don't work in Mastodon right now: https://github.com/tootsuite/mastodon/issues/10861

PR #1219, opened 2019-05-24
Show poll results to a status

sgenoud:
This does not do much, just basics to allows user to see the results of a poll. This is part of #1130.

Capture d'écran 2019-05-24 10 53 01

### DIFF
--- a/src/routes/_components/status/Status.html
+++ b/src/routes/_components/status/Status.html
@@ -31,6 +31,9 @@
   {#if showMedia }
     <StatusMediaAttachments {...params} on:recalculateHeight />
   {/if}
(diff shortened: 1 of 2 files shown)


nolanlawson:
I love it! It just needs some accessibility work for screenreaders, and then I think it's good. (E.g. an `aria-label` that just says exactly what the SVG shows, or some explanatory text with class `sr-only` and then `aria-hidden` on the SVG.) I'll add that and then merge.

Thank you; this is a great start for poll support!

nolanlawson:
my fixes for a11y and some slight style tweaks ^

thanks again!

PR #1220, opened 2019-05-25
fix: tweak poll results style and fix a11y

nolanlawson:
tweaks to the poll results style and fixes a11y issues with it (addendum to #1219)


nolanlawson:
Screenshot from 2019-05-25 08-38-16

PR #1232, opened 2019-05-26
fix: poll for updates to timeago displays

nolanlawson:
When it says "x seconds ago" or "x minutes ago", it should automatically update. We can do this with a simple setInterval that sets the current time in a Svelte store.


nolanlawson:
Turned out this was also a good opportunity to avoid some costly recomputes.

PR #1251, opened 2019-05-28
fix: no need for double reload of SW in Chrome

### LINKED ISSUE #1243: Service Worker can get stuck
In Chrome, I observe that clicking the "reload" button does nothing. It does not update the site until you close and reopen it. (Repro'd in Chrome for Android and in Chrome on Ubuntu.)

In Firefox, it can sometimes get into a state where the site won't load, and even if I clear all the local data and unregister the service worker, I can't get it to load.

Not sure why this is happening.

nolanlawson:
fixes #1243

Sadly it looks like Chrome changed their implementation, which is why in Chrome you currently have to reload twice. I did find a solution that works in both Chrome and Firefox (thank you smoosh.app for boilerplate Service Worker code I can copy-paste).

Although unfortunately it means that the "reload" button in the snackbar toast is now the only way to actually update the Service Worker - browser reload does not work. The worst part about that is that it's not very accessible, because keyboard users can't get to the button if they're in an infinite list. On the bright side, you can just navigate to the settings and then get to the button that way, although it's a bit janky.

I still think this solution is slightly better, though, since it probably confuses Chrome users when they click the "reload" button and nothing happens.

### DIFF
--- a/src/routes/_pages/settings/index.html
+++ b/src/routes/_pages/settings/index.html
@@ -36,7 +36,7 @@
     methods: {
       reload (event) {
         event.preventDefault()
-        document.location.reload(true)
+        location.reload()
       }
     }
   }
--- a/src/routes/_utils/serviceWorkerClient.js
+++ b/src/routes/_utils/serviceWorkerClient.js
@@ -1,17 +1,81 @@
 import { snackbar } from '../_components/snackbar/snackbar'
 
-function onUpdateFound (registration) {
-  const newWorker = registration.installing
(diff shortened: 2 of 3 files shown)


nolanlawson:
There must be a solution to make the browser reload button work the same as the snackbar "reload" button, but I'm kind of stumped as to how to do that.

nolanlawson:
I feel extremely unsatisfied due to the accessibility issues of this PR, but I'm just going to hope that keyboard users can figure out to navigate to a page without an infinite load in order to get to the button. I've already added a region landmark, so screenreader users should be able to access it.

Also, I need to get this in early so I can test it on dev.pinafore.social.

PR #1277, opened 2019-06-07
fix: push subscriptions per instance

### LINKED ISSUE #1274: Push subscription state is shared across instances
At present, Pinafore maintains a single source of truth for the push subscription state, as a LocalStorage object named `store_pushSubscription`:
```json
store_pushSubscription:"{"server_key":"snip","id":"72","endpoint":"https://updates.push.services.mozilla.com/wpush/v2/snip","alerts":{"reblog":true,"mention":true,"follow":true,"favourite":true}}"
````
I think this is incorrect, as I can have multiple instances with different configurations.

nolanlawson:
fixes #1274

We don't have any push subscription tests, so I'll have to test manually. I also made an effort to make the old settings carry over if you're only logged in to one instance.

### DIFF
--- a/src/inline-script/inline-script.js
+++ b/src/inline-script/inline-script.js
@@ -14,7 +14,9 @@ const {
   currentInstance,
   instanceThemes,
   disableCustomScrollbars,
-  enableGrayscale
+  enableGrayscale,
+  pushSubscription,
+  loggedInInstancesInOrder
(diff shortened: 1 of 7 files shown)


nolanlawson:
This isn't a great solution, but I've at least added a notice saying "Note that you can only have push notifications for one instance at a time."

PR #1283, opened 2019-06-17
Fix report dialog not taking comments

### LINKED ISSUE #1282: Report comments are not sent
When reporting a user, anything i type into the additional comments box is never sent with the report

(i love that i reported the same issue in mastodon two years ago tootsuite/mastodon#1671)

codl:
fixes #1282


nolanlawson:
Thank you! Looks like the Travis error is an intermittent failure.

PR #1302, opened 2019-07-06
test: test in quarantine mode

nolanlawson:
Trying to figure out which tests are flaky


nolanlawson:
Did not tell us anything useful

nolanlawson:
Well maybe enabling this will make our tests less flaky by running the flaky ones more than once.

nolanlawson:
Actually because our tests are stateful, it's unlikely this will help. A test can modify them most recent status, then the next test creates a status and checks the most recent one, which was actually created by the last iteration of the test. We should just fix our flaky tests.

PR #1307, opened 2019-07-07
fix: remove useless event argument

nolanlawson:


nolanlawson:
This is wrong, I was mistaken

PR #1313, opened 2019-07-07
test: try not using a timeout when restoring mastodon data

nolanlawson:
Let's see if Mastodon still has race conditions here


nolanlawson:
yep, still need to sleep here

PR #1323, opened 2019-07-08
fix: use pure CSS for focus effect

### LINKED ISSUE #1322: Remove awkward applyFocusStylesToParent helper
There has got to be a way to do what I'm trying to do with pure CSS: https://github.com/nolanlawson/pinafore/blob/9f3c05b12121145851a1c063369046e3306f858e/src/routes/_utils/events.js#L39

nolanlawson:
fixes #1322

This is a weird fix, but I don't really know enough CSS to figure out why this isn't working properly, and I why I need to fix the focus outline on some elements.


nolanlawson:
Amusingly it looks like I was already using this exact same trick for the image/video play/open buttons.

PR #1340, opened 2019-07-18
Use unformatted status content in deleteAndRedraft

### LINKED ISSUE #1339: Use the `text` field of a toot when using delete/redraft
Since Mastodon 2.9.0, [the unformatted content of the deleted status is available in the server reply](https://github.com/tootsuite/mastodon/commit/6d44f2441bf2aa14d11e481d9d2cfe82a74d81ed) when one deletes a toot. This has been done to help clients to implement the delete/redraft feature.

Pinafore could now simply check if there is a field named `text` in the json sent when deleting a toot to fill the redraft modal, removing the need to parse the rendered html content.

It would be beneficial to glitch-soc and pleroma users since when they use extra formatting, some of the html tags are stripped and do not appear in the redraft modal (stuff like ul, ol and blockquote)

Aldarone:
Fix #1339 

(Couldn't do tests because I don't have a working mastodon dev environment)

### DIFF
--- a/src/routes/_actions/delete.js
+++ b/src/routes/_actions/delete.js
@@ -6,9 +6,10 @@ import { deleteStatus as deleteStatusLocally } from './deleteStatuses'
 export async function doDeleteStatus (statusId) {
   let { currentInstance, accessToken } = store.get()
   try {
-    await deleteStatus(currentInstance, accessToken, statusId)
+    let deletedStatus = await deleteStatus(currentInstance, accessToken, statusId)
     deleteStatusLocally(currentInstance, statusId)
     toast.say('Status deleted.')
+    return deletedStatus
   } catch (e) {
     console.error(e)
     toast.say('Unable to delete status: ' + (e.message || ''))
--- a/src/routes/_actions/delete.js
+++ b/src/routes/_actions/delete.js
@@ -13,5 +13,6 @@ export async function doDeleteStatus (statusId) {
   } catch (e) {
     console.error(e)
     toast.say('Unable to delete status: ' + (e.message || ''))
+    return {}
   }
 }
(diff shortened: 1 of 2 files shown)


src/routes/_actions/deleteAndRedraft.js
```diff
@@ -6,10 +6,10 @@ import { store } from '../_store/store'
 export async function deleteAndRedraft (status) {
   let deleteStatusPromise = doDeleteStatus(status.id)
   let dialogPromise = importShowComposeDialog()
-  await deleteStatusPromise
+  let deletedStatus = await deleteStatusPromise
 
   store.setComposeData('dialog', {
-    text: statusHtmlToPlainText(status.content, status.mentions),
+    text: deletedStatus.text || statusHtmlToPlainText(status.content, status.mentions),
```
Aldarone: Maybe I should check if deletedStatus is defined in case deletedStatusPromise fails ?

src/routes/_actions/delete.js
```diff
@@ -6,11 +6,13 @@ import { deleteStatus as deleteStatusLocally } from './deleteStatuses'
...
   try {
-    await deleteStatus(currentInstance, accessToken, statusId)
+    let deletedStatus = await deleteStatus(currentInstance, accessToken, statusId)
     deleteStatusLocally(currentInstance, statusId)
     toast.say('Status deleted.')
+    return deletedStatus
   } catch (e) {
     console.error(e)
     toast.say('Unable to delete status: ' + (e.message || ''))
+    return {}
```
nolanlawson: I'd rather throw the error here so that the compose dialog is not shown

nolanlawson:
Thanks for the PR! These instructions should help get your Mastodon environment set up (on Linux/Mac): https://github.com/nolanlawson/pinafore/blob/master/CONTRIBUTING.md#running-integration-tests

nolanlawson:
I tested it with the error throwing, and it correctly deletes the status but does not show the compose dialog (e.g. when offline).

PR #1341, opened 2019-07-19
Refresh the main stream on click on home

sgenoud:
This is trying to solve the same issue as #1146, but in a simpler manner.

When clicking / tapping on the home button, while in that screen, it restarts the main stream. It is less discoverable than a pull to refresh, but works and is out of the way.

### DIFF
--- a/src/routes/_components/NavItem.html
+++ b/src/routes/_components/NavItem.html
@@ -109,7 +109,6 @@
   import { mark, stop } from '../_utils/marks'
   import { doubleRAF } from '../_utils/doubleRAF'
   import { scrollToTop } from '../_utils/scrollToTop'
-  import { refreshStream } from '../_store/observers/instanceObservers.js'
 
   export default {
     oncreate () {
@@ -173,7 +172,8 @@
         }
 
         if (page === 'home') {
-          refreshStream()
+          // As the array is new, it will just trigger a new refresh
+          store.set({ currentInstanceRefresh: [] })
         }
 
         if (scrollToTop(/* smooth */ true)) {
--- a/src/routes/_components/NavItem.html
+++ b/src/routes/_components/NavItem.html
@@ -109,6 +109,7 @@
   import { mark, stop } from '../_utils/marks'
   import { doubleRAF } from '../_utils/doubleRAF'
   import { scrollToTop } from '../_utils/scrollToTop'
+  import { refreshStream } from '../_store/observers/instanceObservers.js'
 
   export default {
     oncreate () {
@@ -166,10 +167,15 @@
     },
     methods: {
       onClick (e) {
-        let { selected } = this.get()
+        let { selected, page } = this.get()
         if (!selected) {
           return
         }
+
+        if (page === 'home') {
+          refreshStream()
+        }
+
         if (scrollToTop(/* smooth */ true)) {
           e.preventDefault()
           e.stopPropagation()
--- a/src/routes/_store/observers/instanceObservers.js
+++ b/src/routes/_store/observers/instanceObservers.js
@@ -114,23 +114,23 @@ export function instanceObservers () {
 
     scheduleIdleTask(() => refreshInstanceDataAndStream(store, currentInstance))
   })
-}
 
-export function refreshStream () {
-  if (!process.browser) {
-    return
-  }
-  if (currentInstanceStream) {
-    currentInstanceStream.close()
-    currentInstanceStream = null
-    if (process.env.NODE_ENV !== 'production') {
-      window.currentInstanceStream = null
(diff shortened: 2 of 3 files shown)


nolanlawson:
I'm okay with this solution, but I have a few questions:

1. Why just the home screen? Couldn't we do this for every screen?
2. Shouldn't this only work when you're scrolled to the top? Typically tapping the button scrolls us to the top. (Useful on non-iOS platforms where you can't easily scroll to the top.)
3. There seems to be some circular dependency errors in Webpack; it may be worth moving `refreshStream()` into `actions`, then having it call `emit()` from `eventBus.js`, then have the `instanceObservers` listen for an `on()` event from the event bus.

sgenoud:
> Why just the home screen? Couldn't we do this for every screen?

We could, but it is not nececessary. The other streams are stopped and started when as you close and open them. It is not possible to do it for the main stream, as it drives not only the home, but notifications as well (this is what I remember from last May...)

> Shouldn't this only work when you're scrolled to the top? Typically tapping the button scrolls us to the top. (Useful on non-iOS platforms where you can't easily scroll to the top.)

It seemed like a good compromise (as it is not very discoverable). People will click on this to scroll and realise that it fixes their loading state (?). It is my assumption as least.

sgenoud:
>     3\. There seems to be some circular dependency errors in Webpack; it may be worth moving `refreshStream()` into `actions`, then having it call `emit()` from `eventBus.js`, then have the `instanceObservers` listen for an `on()` event from the event bus.

Not sure it is possible to do what you propose (as far as I understand). The event bus uses the component lifecycle events to manage the listeners.
But the instanceObserver is outside of any component - so we might run into weird bugs and some memory leaks.

Let me try to find another solution (and if you did not hear from me soon I decided that an apéro was better for me than trying to solve that issue 😉 today).

nolanlawson:
Ah yes, I forgot that the home and notifications streams are treated specially.

I'd also really like to just fix the problems with the streams - as in, why does it sometimes not refresh? Why does it sometimes lose statuses/notifications? I probably need to handle offline/frozen events better.

nolanlawson:
I need to think about this, as I was kind of hoping we could do a pull-to-refresh using pointer events, and/or just fix the issue so you don't need to refresh manually in the first place.

sgenoud:
> I'd also really like to just fix the problems with the streams - as in, why does it sometimes not refresh? Why does it sometimes lose statuses/notifications? I probably need to handle offline/frozen events better.

Have you thought of using the server side events API instead of websockets? I heard that, as long as you don't want two way communication, it is better than websockets. Generally the point made was that the reconnection logic is all handled by the browser and you do not have to build it yourself.

nolanlawson:
I did try SSE; unfortunately it didn't work in Safari. So I went with websockets because that's what the Mastodon frontend uses.

I'm not sure though that SSE really solves the problem with websockets. Let me try to summarize the issue here:

1. You're online, you get toots A B C
2. You go offline or the tab is frozen
3. Toots D, E, F are posted to the server
4.  You go back online
5. Toots G, H, I are posted to the server
6. The client has to realize that it's been offline and ask the server for toots D, E, F so there isn't a "gap" in the timeline

This is supposed to be handled by [this code](https://github.com/nolanlawson/pinafore/blob/2d8b61e58963a8729d2a1093d62166f0fc24cf0d/src/routes/_store/observers/instanceObservers.js#L88-L97) but in practice it seems like sometimes the websocket loses the connection and then never reconnects, and never tries to fill the gap. It's hard to reproduce, so I don't know exactly what causes it.

nolanlawson:
The more I think about "tap the nav button to refresh" the more unsure I am about it:

1. It's not very discoverable, and it's odd that the button should do two things (scroll up and refresh)
2. The browser already has a refresh button – UAs that don't provide this are wrong and should provide a better way to refresh.
3. I would like to fix the "streaming gap" problem in a more generic way so that you don't need to refresh manually. I will try to find time to work on this.
4. Barring that, I would be okay with a pull-to-refresh (#1146) if it used pointer events and didn't interfere with scrolling.

sgenoud:
> I did try SSE; unfortunately it didn't work in Safari. So I went with websockets because that's what the Mastodon frontend uses.

Interesting. What was the issue with safari?

I was thinking if SSE for the issue where the sockets seem to be stuck in the browser (not the missing messages) - as it might behave better. But it seems not 😞. 

I’ll try to find a better solution as well!

nolanlawson:
> Interesting. What was the issue with safari?

I can't recall exactly, but it didn't work. :disappointed: Also SSE aren't supported in Edge yet: https://caniuse.com/#feat=eventsource

PR #1348, opened 2019-07-21
Fix the style of blockquotes, ul and li in toots

### LINKED ISSUE #1324: Markdown formatted toots are not properly styled
Some ActivityPub implementations allow users to send markdown-formatted toots. glitch-soc and pleroma don't alter this formatting and send the HTML-formatted content to the clients. Mastodon have a whitelist of HTML tags and strips everything else.

When using Pinafore with a glitch-soc account, HTML tags are not properly formatted.

For example this one is an unordered list : 
Screenshot_2019-07-08 ltch fr · Home

And this one have a blockquote : 
Screenshot_2019-07-08 ltch fr · Profile

Bold and italic are ok.

Aldarone:
Fix #1324 

Before:

Screenshot_2019-07-21 ltch fr · Status(1)

After:

Screenshot_2019-07-21 ltch fr · Status

### DIFF
--- a/src/routes/_components/status/StatusContent.html
+++ b/src/routes/_components/status/StatusContent.html
@@ -21,15 +21,31 @@
     display: block;
   }
 
-  :global(.status-content p, .status-content blockquote, .status-content ul, .status-content ol) {
+  :global(.status-content li) {
+    font-size: 1.3em;
+  }
+
+  :global(
+      .status-content p,
+      .status-content blockquote,
+      .status-content ul,
+      .status-content ol) {
     margin: 0 0 20px;
   }
 
-  :global(.status-content p:first-child) {
+  :global(
+      .status-content p:first-child,
+      .status-content blockquote:first-child,
+      .status-content ul:first-child,
+      .status-content ol:first-child) {
     margin: 0 0 20px;
   }
 
-  :global(.status-content :last-child) {
+  :global(
+      .status-content p:last-child,
+      .status-content blockquote:last-child,
+      .status-content ul:last-child,
+      .status-content ol:last-child) {
     margin: 0;
   }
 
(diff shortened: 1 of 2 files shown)


src/routes/_components/status/StatusContent.html
```diff
@@ -21,18 +21,29 @@
     display: block;
   }
 
-  :global(.status-content p) {
+  :global(.status-content p, .status-content blockquote, .status-content ul, .status-content ol) {
     margin: 0 0 20px;
   }
 
   :global(.status-content p:first-child) {
```
nolanlawson: Should this just be `:first-child` if we're removing the `p` below?

src/scss/global.scss
```diff
@@ -162,7 +162,7 @@ button, .button {
   }
 }
 
-p, label, input {
+p, label, input, li {
```
nolanlawson: I'm a bit nervous about changing this global style, as it might affect stuff elsewhere. Can we just do a targeted fix for `.status-content li`?

nolanlawson: Just checked, and indeed if you look at the "All"/"Mentions" text on the /notifications page, it's larger now because of this change. Unfortunately global CSS makes for unpredictable changes; I probably should not have put this in there in the first place...

Aldarone:
I agree with your comments (I just didn't dare to add specific rules everywhere without your feedback)

I'll fix this tomorrow when I get my hands back on my laptop (unless you wanna do it before)

nolanlawson:
OK, I fixed the style. I had to actually make some JS changes because I was wrapping content in `<p>` tags if it didn't have a `<p>` (to handle Pleroma/GNU Social), which meant I was adding extra `<p>` tags around Glitch content which contained `<blockquote`/`<ul>`/`<li>`.

I'll make my changes as a separate PR to avoid blocking this. Thanks for your help!

PR #1352, opened 2019-07-22
perf: make inline script smaller

nolanlawson:
[Large inline scripts should be avoided](https://v8.dev/blog/cost-of-javascript-2019). This reduces the size of the inline script from 2817 bytes to 2419 bytes. (~400 byte reduction)

We may want to avoid merging until https://github.com/rollup/rollup-plugin-inject/pull/23 is fixed, though, since there is kind of an annoying warning message when building.

### DIFF
--- a/bin/build-inline-script.js
+++ b/bin/build-inline-script.js
@@ -5,6 +5,7 @@ import path from 'path'
 import { rollup } from 'rollup'
 import { terser } from 'rollup-plugin-terser'
 import replace from 'rollup-plugin-replace'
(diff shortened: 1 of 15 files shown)


nolanlawson:
Removing rollup-plugin-inject for now, size is still down to 2589 which is pretty good for now

nolanlawson:
I'm nervous about making so many changes for such a small benefit, so I'm going to scale this back a bit

nolanlawson:
The targeted fix ends up with 2772 bytes, there's almost no point in doing this

PR #1360, opened 2019-07-24
Refresh the main streams when on page focus

sgenoud:
It would be better to use the [resume page lifecycle event](https://developers.google.com/web/updates/2018/07/page-lifecycle-api), but Safari does not support it yet.

This is yet another attempts at my quest to make sure the main stream are listened to on iOS PWA. This is far from being perfect, as the stream will be refreshed more than necessary in standard browsers (every time the pinafore tab gains the focus). 
On the other hand, we will not be more certain that the main stream is actually working now.

### DIFF
--- a/src/routes/_store/observers/instanceObservers.js
+++ b/src/routes/_store/observers/instanceObservers.js
@@ -114,4 +114,23 @@ export function instanceObservers () {
 
     scheduleIdleTask(() => refreshInstanceDataAndStream(store, currentInstance))
   })
+
+  store.observe('pageVisibilityHidden', async (isHidden) => {
+    if (!process.browser || isHidden) {
+      return
+    }
+    if (currentInstanceStream) {
+      currentInstanceStream.close()
+      currentInstanceStream = null
+      if (process.env.NODE_ENV !== 'production') {
+        window.currentInstanceStream = null
+      }
+    }
+    const currentInstance = store.get().currentInstance
+    if (!currentInstance) {
+      return
+    }
+
+    scheduleIdleTask(() => refreshInstanceDataAndStream(store, store.get().currentInstance))
+  })
 }


nolanlawson:
Thanks! Does [my usage of lifecycle.js](https://github.com/nolanlawson/pinafore/blob/4d098d6b46e63a995e07af6f0c336d9b8d3dab0e/src/routes/_api/stream/TimelineStream.js#L65) not already handle this? Is the issue that the polyfill doesn't correctly support iOS? Or is the issue that I'm listening for freeze/unfreeze events when I should listen for pagehide/pageshow events?

Note that I also very recently updated this code (#1349), so if you're testing on `pinafore.social` rather than `dev.pinafore.social`, you may be seeing the old behavior.

nolanlawson:
Actually it just occurred to me that freeze/unfreeze is irrelevant here, because that logic pauses the stream entirely and then unpause when it's unfrozen, which is not what we want for pageshow/pagehide because then you wouldn't get notifications while the tab is hidden.

In that case, if we're going to refresh the stream on pageshow, then I think we should do the following:

- move this logic to `src/routes/_actions/stream/streaming.js` and ensure that it fires a `reconnect` event when reconnecting, so that the streaming gap logic is run.
- use `lifecycle.js` for consistency's sake. In `streaming.js` I already hook into it to listen for freeze/unfreeze events, so we just need to listen for pageshow as well.

Does that make sense?

sgenoud:
> Thanks! Does my usage of lifecycle.js not already handle this? Is the issue that the polyfill doesn't correctly support iOS? Or is the issue that I'm listening for freeze/unfreeze events when I should listen for pagehide/pageshow events?

I completely missed this! I did not know that it was possible to polyfill this whole behaviour on iOS. 

> In that case, if we're going to refresh the stream on pageshow

I think I see what you mean here (replace the logic that pauses now the websocket with one that refreshes it), right?

sgenoud:
Mmmm, actually, your lifecycle implementation should just work. In my attempt I was just going at it with a blunter approach, using the pagechanged as a proxy for unfreeze. But the exiting from frozen state should just give back life to the websocket client.

Nice implementation, btw.

I will keep testing with `dev.pinafore` and try to add the refresh on pageshow if you changes did not fix my issue.

nolanlawson:
Cool! Okay, let's try out the new approach and see if it fixes iOS Safari. If not, let's go with your approach and do a refresh on `pageshow` (as described in https://github.com/nolanlawson/pinafore/pull/1360#issuecomment-514677307).

sgenoud:
It looks like your changes fixed the issues I was seeing. Thanks!

I have seen that it adds the new elements at the top automatically when coming back from a frozen state  (and not the button to show more). Is that on purpose?

nolanlawson:
@sgenoud Great to hear! Yes, it should do that if you're scrolled to the top. If you're not scrolled to the top, then that's a bug.

PR #1361, opened 2019-07-24
refactor: move all UA-sniffing to its own file

nolanlawson:
I like having this all in one place so we can maybe remove it one day


nolanlawson:
I'm uncomfortable with the `terser` hacks required to make this properly remove dead code (see https://github.com/terser-js/terser/issues/405)

PR #1370, opened 2019-08-03
chore: update standard to v13

nolanlawson:
Another stab at fixing #1333.

### DIFF
--- a/src/routes/_store/storeLite.js
+++ b/src/routes/_store/storeLite.js
@@ -19,7 +19,7 @@ export const storeLite = {
 
   set (obj) {
     if (hasLocalStorage) {
-      for (let [key, value] of Object.entries(obj)) {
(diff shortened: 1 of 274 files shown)


src/routes/_store/storeLite.js
```diff
@@ -19,7 +19,7 @@ export const storeLite = {
 
   set (obj) {
     if (hasLocalStorage) {
-      for (let [key, value] of Object.entries(obj)) {
+      for (const [key, value] of Object.entries(obj)) {
```
saper: This might be causing https://github.com/nolanlawson/pinafore/issues/1565

nolanlawson: See https://stackoverflow.com/a/39045496, it's due to old Firefox

PR #1371, opened 2019-08-04
fix: improve focal points draggable style/perf

nolanlawson:
Some modest style improvements:

- CSS style for the drag state of the indicator
- CSS style for the hover state of the indicator

Perf/functional improvements:

- Throttle all DOM updates so they only happen once
- Fix an issue where the drag animation would happen twice because of store updates syncing back

Beyond this, the only way I can see to improve perf in Firefox for Android and iOS is to not use the Pointer Events polyfill but rather native touch events. I profiled in Firefox for Android, and most of the time spent is just in the polyfill itself. (Oddly Firefox supports Pointer Events on desktop, but not Android.)


nolanlawson:
OK, I went ahead and fixed all the things. Tested browsers:

- iOS Safari 12
- Firefox for Android
- Chrome for Android
- Chrome Ubuntu
- Firefox Ubuntu
- Epiphany Ubuntu
- Safari Mac

Issue #1365, opened 2019-07-30
Click notification should open app and corresponding post


pwFoo:
If I click notification it doesn't open the app in chrome for android. Maybe it works with a open tab in background, but it should open app ("add to homescreen") or browser if no tab / app in background?

nolanlawson:
Hmmm this is interesting. So you choose "add to homescreen" in Chrome, and then you enabled notifications, but then when you click the notification, it opens the app? Or the browser?

Also, is this regular Chrome with Google Play Services, or is it LineageOS with Chromium or something similar?

pwFoo:
- Chrome for Android (PlayStore)
- Android 9 (OxýgenOS 9.0.7)

If I click the notification nothing happens. Notification is still shown.
I don't know if it happens with desktop pc (Win10, Chrome). Most of the time I use it mobile (OnePlus 5).

pwFoo:
Updated pinafore to latest / 1.11.6.

pwFoo:
Hi @nolanlawson,
something different with the release v1.11.7 because my docker image build fails?
```
Step 7/19 : RUN VERSION=${VERSION:-$(curl --silent $REPO/releases/latest | sed 's#.*tag/\(.*\)\".*#\1#')};      echo $VERSION;  git checkout $VERSION
 ---> Running in 1f6a1ba179ac
v1.11.7
error: pathspec 'v1.11.7' did not match any file(s) known to git
The command '/bin/sh -c VERSION=${VERSION:-$(curl --silent $REPO/releases/latest | sed 's#.*tag/\(.*\)\".*#\1#')};      echo $VERSION;  git checkout $VERSION' returned a non-zero code: 1
```

Looks like v1.11.7 isn't available for checkout? v1.11.7 works fine with that line. Also tried to set VERSION to v1.11.6 and 1.11.7, but only the older one works.

pwFoo:
Fixed!
Looks like a docker cache problem...

pwFoo:
Hi @nolanlawson,
click android notification still not opens pinafore app. Seems not fixed? Could you please reopen the issue and check it? Maybe an chrome for android / mobile problem or something wrong with the manifest.json / app url?

PR #1378, opened 2019-08-07
fix(service-worker): Open notifications if no status

sorin-davidoi:
If the notification couldn't be fetched fallback to displaying all notifications (aligned with how the Mastodon Web UI [handles it](https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/service_worker/web_push_notifications.js#L118)).

Might fix https://github.com/nolanlawson/pinafore/issues/1365.


nolanlawson:
Thanks!

Issue #1179, opened 2019-05-05
Implement blurhash


nolanlawson:
This seems like a nice feature: https://blog.joinmastodon.org/2019/05/improving-support-for-adult-content-on-mastodon/

PR #1381, opened 2019-08-08
feat(media): Blurhash

sorin-davidoi:
Todo:
 - [x] Handle medias without blurhash
 - [x] Look into performance (maybe cache blurhash -> base64 computation)
 - [x] Implement for videos
 - [x] Fix integration tests
 - [x] Option to disable usage

image

Closes https://github.com/nolanlawson/pinafore/issues/1179.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,8 @@
     "mocha": "^6.1.4",
     "now": "^15.7.0",
     "standard": "^13.1.0",
-    "testcafe": "^1.2.1"
+    "testcafe": "^1.2.1",
+    "worker-loader": "^2.0.0"
   },
   "engines": {
     "node": ">= 8"
@@ -152,7 +153,9 @@
       "AbortController",
       "matchMedia",
       "MessageChannel",
-      "ImageData"
+      "ImageData",
+      "OffscreenCanvas",
+      "postMessage"
     ],
     "ignore": [
       "dist",
--- a/package.json
+++ b/package.json
@@ -151,7 +151,8 @@
       "customElements",
       "AbortController",
       "matchMedia",
-      "MessageChannel"
+      "MessageChannel",
+      "ImageData"
     ],
     "ignore": [
       "dist",
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "@webcomponents/custom-elements": "^1.2.4",
     "babel-loader": "^8.0.6",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
+    "blurhash": "^1.1.3",
     "cheerio": "^1.0.0-rc.2",
     "child-process-promise": "^2.2.1",
     "chokidar": "^3.0.1",
--- a/src/routes/_components/status/Media.html
+++ b/src/routes/_components/status/Media.html
@@ -1,12 +1,5 @@
 {#if type === 'video' || type === 'audio'}
-  <button id={elementId}
-          type="button"
-          class="play-video-button focus-after {$largeInlineMedia ? '' : 'fixed-size'} {type === 'audio' ? 'play-audio-button' : ''}"
-          aria-label="Play video: {description}"
-          style="width: {inlineWidth}px; height: {inlineHeight}px;">
-    {#if !blurhash}
-      <PlayVideoIcon />
-    {/if}
+  {#if blurhash}
     {#if type === 'video'}
       <LazyImage
         alt={description}
@@ -20,7 +13,28 @@
         {focus}
       />
     {/if}
-  </button>
+  {:else}
+    <button id={elementId}
+            type="button"
+            class="play-video-button focus-after {$largeInlineMedia ? '' : 'fixed-size'} {type === 'audio' ? 'play-audio-button' : ''}"
+            aria-label="Play video: {description}"
+            style="width: {inlineWidth}px; height: {inlineHeight}px;">
+      <PlayVideoIcon />
+      {#if type === 'video'}
+        <LazyImage
+          alt={description}
+          title={description}
+          src={previewUrl}
+          fallback={oneTransparentPixel}
+          blurhash={blurhash}
+          width={inlineWidth}
+          height={inlineHeight}
+          background="var(--loading-bg)"
+          {focus}
+        />
+      {/if}
+    </button>
+  {/if}
 {:else}
   <button id={elementId}
           type="button"
--- a/src/routes/_components/status/Media.html
+++ b/src/routes/_components/status/Media.html
@@ -4,7 +4,9 @@
           class="play-video-button focus-after {$largeInlineMedia ? '' : 'fixed-size'} {type === 'audio' ? 'play-audio-button' : ''}"
           aria-label="Play video: {description}"
           style="width: {inlineWidth}px; height: {inlineHeight}px;">
-    <PlayVideoIcon />
+    {#if !blurhash}
+      <PlayVideoIcon />
+    {/if}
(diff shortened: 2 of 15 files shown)


src/routes/_utils/blurhash.js
```diff
@@ -0,0 +1,14 @@
+import { decode as decodeBlurHash } from 'blurhash'
+
+let canvas
+
+export function decode (blurhash) {
+  const pixels = decodeBlurHash(blurhash, 320, 320)
+
+  if (pixels) {
+    canvas = canvas || document.createElement('canvas')
```
sorin-davidoi: Mastodon renders the canvases directly but I though that it's too much overhead to create and destroy that media canvases so I went with this approach.

src/routes/_utils/blurhash.js
```diff
@@ -0,0 +1,14 @@
+import { decode as decodeBlurHash } from 'blurhash'
+
+let canvas
+
+export function decode (blurhash) {
+  const pixels = decodeBlurHash(blurhash, 320, 320)
```
sorin-davidoi: Not sure what the best values would be here.

package.json
```diff
@@ -48,6 +48,7 @@
     "@webcomponents/custom-elements": "^1.2.4",
     "babel-loader": "^8.0.6",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
+    "blurhash": "^1.1.3",
```
sorin-davidoi: We could try https://github.com/fpapado/blurhash-rust-wasm if performance turns out to be an issue.

src/routes/_utils/blurhash.js
```diff
@@ -0,0 +1,23 @@
+import { decode as decodeBlurHash } from 'blurhash'
+import { mark, stop } from './marks'
+
+const RESOLUTION = 32
```
sorin-davidoi: We could maybe lower this based on some heuristics (e.g. `navigator.hardwareConcurrency`)? Setting it to 1 will basically extract the average color, so maybe we can do that initially and repeat with 32 inside a `requestIdleCallback`?

src/routes/_utils/blurhash.js
sorin-davidoi: Profile sample with 6x CPU throttling.
image

src/routes/_utils/blurhash.js
nolanlawson: I'm not sure what the limiting factor is for blurhash, but yeah for now a hard-coded value is probably fine.

src/routes/_utils/blurhash.js
nolanlawson: I wonder if `toBlob()` with a Blob URL may actually be more performant here than a data URL. I'll play around with it. :)

nolanlawson: Hm, so the problem with `toBlob()` is it's asynchronous, meaning that there might be a flash of unhidden sensistive media unless we're careful about doing the blurhashing in advance. I might keep noodling on this after merging the PR.

nolanlawson:
Thanks very much for this PR!! I made some minor refactoring changes. At this point, my only major concerns are:

1. This doesn't work with "show large inline images." Although honestly maybe it's time to just remove that option; I'm not sure how many people are using it, given that @sgenoud's new image system is so much better than the old one.
2. In the settings, "Use blurhash for previews" is now the only option that is checked by default. Up to now I have tried to prefer "false by default" to reduce cognitive load on users (i.e. so they can remember what the defaults are). I admit though that it's hard to think of a negation for this feature that wouldn't be awkward. ("Hide blurhash"?)
3. Yeah the performance of blurhashing is a bit concerning, especially when I turn on 6x slowdown in Chrome (up to 100ms, dominates the time spent rendering a status overall). I think long term though, we could move toward a webworker-based solution or `toBlob()` solution, but that would probably require doing some initial asynchronous work before rendering (e.g. in `VirtualListLazyItem.html`). I've been meaning to do that for some other stuff anyway. But that can wait for a future PR.

For now I'm mostly concerned about # 1; maybe it's time for me to retire the "show large inline images system." Or we could implement blurhash for it.

nolanlawson:
Created a poll to see if anyone is using this setting: https://mastodon.technology/@pinafore/102593964013067092

sorin-davidoi:
Pushed a tentative fix for the overflow issue when "show large inline images" is enabled.

image

nolanlawson:
Awesome! In that case I'd say remaining tasks are:

- [ ] Negate the setting somehow... maybe "Show a plain gray color for sensitive media" ?
- [ ] Fix the performance issues... maybe indeed `blurhash-rust-wasm` is the best fix here

sorin-davidoi:
Can't seem to get `blurhash-rust-wasm` to play along (getting `Error: Cannot find module 'blurhash-wasm'`).

sorin-davidoi:
Pushed a commit where the computations are moved to a Web Worker:
 - if `OffscreenCanvas` is supported the main thread sends the encoded string and gets back an `URL`
 - otherwise it sends the encoded string and gets back an `ImageData`, which it then renders using the canvas

Some things still need to be addresses though:
 - the Web Worker is created lazily (when the first blurhash needs to be computed) - this introduces some delay, we should call `init` as soon as we know that blurhashes are not disabled (when the store is created?)
 - some caching for URLs should probably be added, especially since we currently don't ever call `URL.revokeObjectURL`

src/routes/_components/status/Media.html
```diff
@@ -141,6 +166,7 @@
     },
     data: () => ({
       oneTransparentPixel: ONE_TRANSPARENT_PIXEL,
+      decodedBlurhash: ONE_TRANSPARENT_PIXEL,
```
sorin-davidoi: `ONE_TRANSPARENT_PIXEL` is used while the async computation is pending.

src/routes/_utils/blurhash.js
```diff
@@ -0,0 +1,49 @@
+import BlurhashWorker from 'worker-loader!../_workers/blurhash' // eslint-disable-line
```
sorin-davidoi: I couldn't get `worker-loader` working when added to `client.config.js` for some reason. Standard complains about the inline use of the loader, disabled it for now.

sorin-davidoi:
Added caching, but we are still leaking memory due to never calling `revokeObjectURL` - not sure when would be the best time to call it.

sorin-davidoi:
There are still some overflow issues caused by `padding-bottom: 56.25%`:

image

src/routes/_components/status/Media.html
nolanlawson: Good workaround!

src/routes/_utils/blurhash.js
nolanlawson: That's fine; I don't really care whether the `worker-loader` is referenced in the Webpack config or in the import.

nolanlawson:
Fantastic work, this looks great to me. I only have some minor issues remaining:

- The eye icon and "sensitive media" text are a bit hard to read; I think the font should be lighter when blurhash is enabled
- It would be nice to do the worker operations at the same time that the content is fetched asynchronously from IDB so that there isn't a "pop" when the color changes from gray to the blurhash (i.e. delay rendering the toot until the blurhash is ready... it's a tradeoff I'd prefer to make in favor of less visually jarring rendering).

Both of these things can be fixed later, though; overall I think this is a fantastic PR and we should just get it in master and fix the minor things later. Thanks for the excellent work!!

PR #1398, opened 2019-08-18
fix(workers/blurhash): Don't use OffscreenCanvas

sorin-davidoi:
I propose we leave the implementation in place so we can easily test later on if the performance improved.


nolanlawson:
Yup, this seems like the prudent thing for now. I've opened a bug on Chromium; hopefully they'll take a look at it. http://crbug.com/995099

PR #1401, opened 2019-08-18
feat(settings): Increase clickable area

sorin-davidoi:
This increases the clickable area for the `input` while also making sure that there is no "dead" space between the list items. It also gets rid of having to keep the ids in sync with the `for` attribute.

# Before
image
 (the `span` was added to the DOM to illustrate the area)

# After
image


nolanlawson:
Thanks!

PR #1403, opened 2019-08-18
perf: call revokeObjectURL() on stale blurhash blob URLs

### LINKED ISSUE #1402: Call revokeObjectURL on stale blurhash URLs
https://github.com/nolanlawson/pinafore/pull/1381

This will avoid a memory leak. (In theory – I'm not sure if browsers have some way to do this automatically or not. Although we shouldn't take the chance.)

nolanlawson:
fixes #1402

I forked quick-lru to do this. The added benefit is that I could remove a bunch of unused code from that module.

### DIFF
--- a/src/routes/_utils/blurhash.js
+++ b/src/routes/_utils/blurhash.js
@@ -1,13 +1,20 @@
 import BlurhashWorker from 'worker-loader!../_workers/blurhash' // eslint-disable-line
 import PromiseWorker from 'promise-worker'
 import { BLURHASH_RESOLUTION as RESOLUTION } from '../_static/blurhash'
-import QuickLRU from 'quick-lru'
+import { QuickLRU } from '../_thirdparty/quick-lru/quick-lru'
 
-const CACHE = new QuickLRU({ maxSize: 100 })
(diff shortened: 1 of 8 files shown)


src/routes/_utils/blurhash.js
```diff
@@ -5,6 +5,7 @@ import { QuickLRU } from '../_thirdparty/quick-lru/quick-lru'
 
 // A timeline will typically show 20-30 articles at once in the virtual list. The maximum number
 // of sensitive images per article is 4. 30*4=120, so this is a very conservative number.
+// Blurhash blobs seem to range from ~1.2-2kB, so this cache could grow to about 2kB*150=300kB max.
```
sorin-davidoi: Make it can be force-evicted (and the URLs revoked) when [`memorypressure`](https://www.chromestatus.com/features/5719614029824000) is emmited?

nolanlawson: That seems reasonable yeah, but `memorypressure` is apparently not implemented in any browser, even Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=708752

Issue #1404, opened 2019-08-18
Handling of grouped images with blurhash is broken


sorin-davidoi:
Mentioned it in https://github.com/nolanlawson/pinafore/pull/1381#issuecomment-521699935 but forgot to open issue. Can be replicated with https://witches.live/@mal/102639905266205838.

![image](https://user-images.githubusercontent.com/2109702/63230206-d248ee80-c209-11e9-96fc-a37ae99da12d.png)

PR #1409, opened 2019-08-19
fix(components/media): Handling of grouped videos

sorin-davidoi:
Dislike using extra markup to solve style issues but this is the simplest fix.

Fixes https://github.com/nolanlawson/pinafore/issues/1404.


nolanlawson:
I'm okay with this solution! Sometimes it's hard to avoid wrapper divs :man_shrugging: 

Thanks a lot!

PR #1414, opened 2019-08-20
perf: be more consistent about compose input scheduling

nolanlawson:
Follow-up to #1413. Try to be more consistent about using requestIdleCallback everywhere, and to do it the same way across all the input elements in the ComposeBox.


nolanlawson:
Travis CI integration not working apparently: https://travis-ci.org/nolanlawson/pinafore/builds/574399948

PR #1422, opened 2019-08-22
fix: ensure "disable blurhash" setting is honored

### LINKED ISSUE #1421: Disabling blurhash in settings is broken
Regressed by #1418, didn't notice until just now.

Screenshot from 2019-08-21 22-42-53

nolanlawson:
fixes #1421

1. Make blurhash the logical default - if it's sensitive, then it's blurhash
2. If the blurhash is missing (e.g. Pleroma) or disabled (e.g. in the settings), then convert `--mask-bg` into a data URL and treat that as the blurhash

This slightly changes the way things look when blurhash is disabled (e.g. instead of a solid color, you'll see white gaps between multiple images), but I think that's fine.

Screenshot from 2019-08-22 08-11-04

### DIFF
--- a/src/routes/_components/status/Media.html
+++ b/src/routes/_components/status/Media.html
@@ -1,4 +1,4 @@
-{#if (type === 'video' && !blurhash) || type === 'audio'}
+{#if type === 'video' || type === 'audio'}
     <button id={elementId}
             type="button"
             class="play-video-button focus-after {$largeInlineMedia ? '' : 'fixed-size'} {type === 'audio' ? 'play-audio-button' : ''}"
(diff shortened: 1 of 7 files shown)


src/routes/_components/status/Media.html
```diff
@@ -1,4 +1,4 @@
-{#if (type === 'video' && !blurhash) || type === 'audio'}
+{#if type === 'video' || type === 'audio'}
```
sgenoud: you probably still want to have the case with or without blurhash different here. You do not want to have the play triangle to show in blurhash mode (I guess).

nolanlawson: Ah, I noticed that, but didn't think it was a big deal. You're right though, that triangle is weird-looking.

nolanlawson: OK, I fixed it. BTW it turns out that audio can be marked as sensitive too, although it's kind of weird to me, but I decided to treat it the same as video.

PR #1433, opened 2019-08-25
feat: add OCR for image descriptions using tesseract.js

nolanlawson:
Another feature shamelessly stolen from Mastodon.

I'm not super excited or satisfied with this so far, but it's a good accessibility feature so I think it's worth having.

Problems:

- tesseract doesn't play nice with CSP, this required hacks to work around https://github.com/naptha/tesseract.js/pull/322
- tesseract has a bug with broken/failed image URLs https://github.com/naptha/tesseract.js/issues/325
- tesseract doesn't seem to support WASM yet; I have to use the `wasm.js` file which is ~1MB bigger
- some instance admins may not have enabled CORS for media assets (I hadn't). If so this feature won't work
- it's a big dependency and really awkward to manage like 3 separate files using file-loader and other weird hacks

Overally though I'm willing to move forward with this; hopefully tesseract will get better when it's out of alpha/beta.


nolanlawson:
Screenshot from 2019-08-25 15-48-43

Issue #1347, opened 2019-07-20
Revoking access should clear data


nolanlawson:
If Pinafore is revoked access to an instance, Pinafore should recognize that and just log out and clear data.

pwFoo:
Yes, at the moment it looks like a internet connection problem if access is revoked.

PR #1446, opened 2019-08-28
fix(actions/instances): Clear data when token revoked

sorin-davidoi:
Only handles `401` thrown by `/api/v1/accounts/verify_credentials` - ideally all endpoints should be covered. There are some concerns with doing that:
 - What if the user typed something in the compose box and we log them out when they submit the form?
 - Should we add `.catch` to all requests or only once in a global place (e.g. `ajax.js`)? If we do it at a lower level we don't have access to the instance name.
 - Since we launch multiple requests in parallel, the same toast might get displayed multiple times - ideally it should only happen once.

Closes https://github.com/nolanlawson/pinafore/issues/1347.

### DIFF
--- a/src/routes/_actions/instances.js
+++ b/src/routes/_actions/instances.js
@@ -31,7 +31,7 @@ export function switchToInstance (instanceName) {
   switchToTheme(instanceThemes[instanceName], enableGrayscale)
 }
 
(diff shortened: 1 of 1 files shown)


nolanlawson:
Thank you!

Issue #812, opened 2018-12-15
Don't fetch the emoji/list on every page load


nolanlawson:
It seems silly to ask the instance "hey what are your custom emojis and lists" on every single page load. We can delay this call until the emoji panel is opened, or until the "community" column is opened in the case of the latter.

nolanlawson:
Interestingly, it looks like the Mastodon frontend also does it.

PR #1448, opened 2019-08-28
refactor(store/observers/instance): Don't fetch lists eagerly

sorin-davidoi:
They are already fetched on-demand when the community column is loaded.

Part of https://github.com/nolanlawson/pinafore/issues/812.


nolanlawson:
Thanks!

PR #1450, opened 2019-08-30
perf: fetch lists from database on page load

nolanlawson:
follow-up to #1448, I would like for there not to be a flash on the /community page where the lists suddenly load in, but I would still like to avoid the network request if possible. So when the page first loads, we can load the lists from the database and only fetch them from the network if not in the cache.

also related to #812


nolanlawson:
Incidentally, this also fixed an issue where if you pinned a list to the nav bar, it wouldn't show the name of the list until you clicked the "community" button.

PR #1451, opened 2019-08-30
perf: don't fetch custom emoji on every page load

### LINKED ISSUE #812: Don't fetch the emoji/list on every page load

nolanlawson:
fixes #812

follow-up to #1448 and #1450

### DIFF
--- a/src/routes/_actions/emoji.js
+++ b/src/routes/_actions/emoji.js
@@ -1,10 +1,13 @@
-import { cacheFirstUpdateAfter } from '../_utils/sync'
+import {
+  cacheFirstUpdateAfter,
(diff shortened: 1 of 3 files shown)


nolanlawson:
With this PR we only ever update custom emoji if you actually click the emoji dialog button - not when auto-completing emoji in the text box. And even then, we only do it lazily, so you might have to click the button twice.

I think this is reasonable, as an instance's custom emoji probably don't get updated that frequently. Also I assume that if people want the latest custom emoji, they will refresh and hit the emoji button until it's fresh.

PR #1453, opened 2019-08-30
fix: h2 in media edit dialog should be h1

nolanlawson:
The dialog is in its own document, so the heading level should start
back at h1.


nolanlawson:
Actually I just realized there is already an h1 in there, so actually the h2 is correct >_<

PR #1472, opened 2019-09-08
fix: allow audio uploads on iOS Safari

### LINKED ISSUE #1461: Allow m4a upload format
Mastodon recently updated the allowed audio types: https://github.com/tootsuite/mastodon/pull/11342

Also we should probably switch from an image icon to a paperclip icon to make it clear you can upload audio.

nolanlawson:
fixes #1461

I decided against changing the camera icon to the paperclip icon for now because I think the paperclip icon looks ugly (it's not square-ish like the other icons, so looks out of place). Also image/video is the 99.5% use case, so it's less confusing to the average user to have a camera icon. I did add a title/aria-label to indicate that audio is accepted, though.

The current Safari solution of using `accept=*/*` to allow audio is not great, but we have to wait for the WebKit bug to get fixed before changing it.

### DIFF
--- a/src/routes/_components/compose/ComposeToolbar.html
+++ b/src/routes/_components/compose/ComposeToolbar.html
@@ -9,7 +9,7 @@
     <IconButton
       className="compose-toolbar-button"
       svgClassName={$uploadingMedia ? 'spin' : ''}
-      label="Add media"
+      label="Add media (images, video, audio)"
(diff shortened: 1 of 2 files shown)


nolanlawson:
Oh yeah also you can now upload multiple files at once, which was an easy fix.

PR #1486, opened 2019-09-15
feat: mark media as sensitive without a CW

### LINKED ISSUE #1297: Cannot mark media as sensitive when a toot has no content warning
I can't find the « mark as sensitive » button for uploaded media but I found people making reference to it.

Perhaps it's aregression, perhaps it's my broken brain ? :confused: 

Screenshot_2019-07-04 ltch fr · Home(1)

nolanlawson:
fixes #1297

With this fix, media can be marked as sensitive whether or not there is a content warning. A checkbox is added to toggle sensitivity, just like the Mastodon UI.

The checkbox will automatically turn itself on and off depending on whether or not there is a CW. But you can flip it to "on" even without a CW.

Note that the Mastodon API does not allow the reverse scenario (a content warning but a non-sensitive media).

### DIFF
--- a/src/routes/_actions/contentWarnings.js
+++ b/src/routes/_actions/contentWarnings.js
@@ -6,6 +6,7 @@ export function toggleContentWarningShown (realm) {
   const newShown = !shown
   store.setComposeData(realm, {
     contentWarning: newShown ? contentWarning : '',
-    contentWarningShown: newShown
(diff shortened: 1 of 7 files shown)


nolanlawson:
Screenshot from 2019-09-15 09-42-51

PR #1491, opened 2019-09-16
fix: remove avatars from tab order / screenreaders

nolanlawson:
These elements are duplicated information (the link and name are right next to them), so they should be hidden from screenreaders and tab order.

I also went ahead and fixed a place where I think I should be using aria-busy instead of aria-hidden, because it's content that's asynchronously loading.

### DIFF
--- a/src/routes/_components/compose/ComposeAuthor.html
+++ b/src/routes/_components/compose/ComposeAuthor.html
@@ -1,14 +1,17 @@
 <a {href}
    rel="prefetch"
    class="compose-box-avatar {loaded ? 'loaded' : 'not-loaded'}"
-   aria-hidden={true}
+   aria-hidden="true"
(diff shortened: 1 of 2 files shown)


nolanlawson:
What's really weird is that VoiceOver on Chrome on macOS does not seem to actually take the avatars out of the screen reader visibility, and they just show up as "link" with no name. But I'm not sure if it's a VoiceOver bug or what.

nolanlawson:
Based on https://stackoverflow.com/q/40659260 it may be a bug in Chrome, so I'll have to test in Safari once I get access to a macOS device again.

nolanlawson:
I'll test before releasing to prod, anyway the aria-hidden was already there

nolanlawson:
It works correctly on VoiceOver in Safari.

PR #1508, opened 2019-09-21
fix: fix shortcut help dialog not keyboard-scrollable

### LINKED ISSUE #1473: keyboard shortcut screen cannot be scrolled by keyboard
I attempt to scroll up and down in the keyboard shortcut help screen (activated by '?') but nothing happens. j,k do not work; up and down arrows do not work; page up and down keys do not work. I had to use the mouse to learn the keyboard shortcuts. Is there a possibility to add keyboard scrolling to this help view? Thanks.

nolanlawson:
fixes #1473


nolanlawson:
When the shortcut info is in a dialog, it's now tab-indexable and shows the scrollbar while focused.

Unfortunately I cannot get the scrollbar to re-show in Chrome because of the way the webkit-specific scrollbar logic is implemented (if you change it once, you can never change it back). But it at least re-shows on Firefox, and at least on all browsers you can scroll correctly.

Screenshot from 2019-09-21 12-01-35

PR #1511, opened 2019-09-22
fix: fix pleroma hashtag links

### LINKED ISSUE #832: Pleroma mentions and hashtags open in new window, not locally
When someone is mentioned in Pleroma, if you click the handle it will take you to their public profile, rather than the local one.

nolanlawson:
fixes #832


nolanlawson:
So mentions are fine; the only issue is hashtag links, which Pleroma marks as `/tag/foo` rather than `/tags/foo`. I hate hard-coding anything, so I'm just going to remove the tags/tag checking from the URL checking logic entirely.

PR #1516, opened 2019-09-22
feat: left/right keys can change columns or focus

nolanlawson:
This is an important accessibility feature for KaiOS (using left/right to change focus, as described in [my blog post](https://nolanlawson.com/2019/09/22/the-joy-and-challenge-of-developing-for-kaios/)), but it could also potentially be used by anyone who wants this feature.

By default, the left/right arrow keys now change the column (or the media in the media dialog). But with this setting checked, left/right changes the focus instead. On KaiOS, this setting is enabled by default and cannot be changed.


nolanlawson:
Broke something, I should probably write some tests for this anyway...

Issue #1512, opened 2019-09-22
[Accessibility] Don't prepopulate aria-activedescendant with a pointer to a hidden list


MarcoZehe:
Due to some [changes in Firefox](https://bugzil.la/686400), a bug was uncovered that both earlier versions of Firefox and current versions of Chrome are masking. When initially constructing the text input, after #129 was fixed, aria-activedescendant already points to the ul with ID ´compose-autosuggest-list-home`. That UL, since nothing has been typed yet, or when no suggestions are available, is inside a div that is aria-hidden="true", and thus also hidden. Firefox is, however, due to a bug I still need to file, creating an accessible for it for some reason now. That accessible is now masking the input field. The result is that I am not able to see what I am typing, and focus is inside a list that is essentially not there.

To fix, it would be best if aria-activedescendant was not set initially, and only gets set if the list is visible and a list item actually needs to be focused due to it being an autocomplete result. If the autocomplete goes away, and the list is hidden, aria-activedescendant should be cleared again. Meaning the attribute be removed.

MarcoZehe:
Just confirmed via the HTML inspector that removing this attribute fixes the problem. So, it shold definitely not be set initially when preparing the textarea for entry. :) Thanks!

nolanlawson:
Thanks for reporting! I should be able to test on Firefox+NVDA to confirm. Perhaps VoiceOver in Safari was also masking this issue.

nolanlawson:
Just to check: it looks like initially, `aria-owns` points to `compose-autosuggest-list-home` (which is indeed inside of an `aria-hidden`), but `aria-activedescendant` is initially set as an empty string. Are you saying that `aria-owns` should not be set if the list is hidden, or that `aria-activedescendant` should be undefined rather than an empty string?

Also, which version of Firefox should I test to repro the bug? Is this only in Nightly?

MarcoZehe:
This is in Nightly and Beta or Dev Edition. aria-activedescendant is always set for me, it is not an empty string when the empty text area gains focus and I explore its properties in HTML inspector. For safety, I think it should be undefined rather than an empty string..

nolanlawson:
Should be fixed by #1513 if I understood correctly! Thanks for your help.

PR #1520, opened 2019-09-23
fix: remove aria-live region for autosuggest

nolanlawson:
After some testing and based on [this suggestion](https://toot.cafe/@marcozehe/102840228729195138) from @MarcoZehe, I think we should just remove the `aria-live` region from the autosuggest. It always felt hacky to me anyway.

I got the original idea from [this implementation](https://haltersweb.github.io/Accessibility/autocomplete.html), but I wonder if it's outdated or wrong. My current implementation seems to cause double utterances in Firefox and some weird garbled text (letter artifacts) as well.

With this new implementation, the results are _perfect_ in Firefox Developer Edition (v70) using NVDA. All other environments I tested have some bugs, but are usable IMO.

- Safari on macOS with VoiceOver: announces that autocomplete results are available, but doesn't announce the first item in the list. Only starts announcing items after you press up and down. Sometimes it gets into a state where it doesn't announce anything, but this only seems to occur when the tab focus and screenreader focus get out of sync, and can be fixed by tabbing out and back in again. I have not found a fix for that.
- Firefox v69 with NVDA: announces that autocomplete results are available, but doesn't announce the first item. Like Safari, you have to press up or down to see the first item
- Chrome with NVDA: only announces the first item. Pressing up or down doesn't announce any new items. But at least it announces the first one.

I tried _not_ swapping out `aria-owns` whenever the owned element becomes `aria-hidden`, but this broke Safari entirely (it stopped speaking any results).

Overall this PR seems to be a slight regression for Chrome+NVDA, no change for Safari+VoiceOver, and a big improvement for Firefox+NVDA. Since I assume NVDA is most commonly used with Firefox, this PR seems like the best implementation I can think of for now.

### DIFF
--- a/src/routes/_components/compose/ComposeAutosuggest.html
+++ b/src/routes/_components/compose/ComposeAutosuggest.html
@@ -8,9 +8,6 @@
     selected={autosuggestSelected}
     {realm}
   />
-  <div class="sr-only" aria-live="assertive">
-    {assertiveAriaText}
-  </div>
 </div>
 <style>
   .compose-autosuggest {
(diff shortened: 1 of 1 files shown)


nolanlawson:
Related: #1512

PR #1538, opened 2019-09-25
perf: lazy-load computations (experimental)

nolanlawson:
I thought it might be neat to try to lazy-load store computations we don't need until you're logged in. It turns out that it only saves ~2KB of JS gzipped (~4KB non-gzipped), and on 6x slowdown maybe reduces one of the tasks from ~40ms to ~15ms. Really seems not worth it, but I'm curious if it even passes the tests.

### DIFF
--- a/src/routes/_components/NavItem.html
+++ b/src/routes/_components/NavItem.html
@@ -144,7 +144,7 @@
         (name === 'notifications' && $hasNotifications) || (name === 'community' && $hasFollowRequests)
       ),
       badgeNumber: ({ name, $numberOfNotifications, $numberOfFollowRequests }) => (
-        (name === 'notifications' && $numberOfNotifications) || (name === 'community' && $numberOfFollowRequests)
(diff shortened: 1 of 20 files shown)


nolanlawson:
Hm, it's kind of a micro-optimization, but it does seem stable.

nolanlawson:
6x slowdown mark/measure cost: 

- timelineComputations 39.92ms
- autosuggestComputations: 16.28ms
- navComputations: 1.69ms
- instanceComputations: 9.13ms

yeah maybe it is slightly worth it, if the risk is low

nolanlawson:
Current total main thread cost at 6x slowdown: 221ms scripting, 59ms rendering

After this fix: 184ms scripting, 62ms rendering

OK yeah, that's a fairly big chunk

nolanlawson:
Even better: if we lazy-load mixins, then we can get down to 77KB from 83.4KB (ungzipped), so 6.4KB saved.

PR #1540, opened 2019-09-25
Add "unfavorite" label.

chexxor:
Just wanted to float this idea past you maintainers.

I saw a filled-in star on a toot, but I couldn't figure out whether I had already favorited it or not. I hovered over it and saw it said "Favorite", so I thought if I click it then it will favorite the toot. It actually unfavorited it.

So let me know what you think!

I didn't test this yet, btw.

### DIFF
--- a/src/routes/_components/status/StatusToolbar.html
+++ b/src/routes/_components/status/StatusToolbar.html
@@ -19,7 +19,7 @@
     ref:reblogIcon
   />
   <IconButton
-    label="Favorite"
+    label="{favoriteLabel}"
     pressable="true"
     pressed={favorited}
     href="#fa-star"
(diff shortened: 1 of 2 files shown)


nolanlawson:
This sounds great to me! Seems like a true bug that we overlooked.

There is a minor syntax error, but I can fix that, no worries. Thank you!

nolanlawson:
OK, I fixed up the syntax issue. Also I noticed that we made the same mistake for Boost/Unboost, so I fixed that that.

Thanks for pointing this out! Amazing that this bug went by unnoticed for so long. It's also an accessibility fix for screenreader or assistive technology users. :)

chexxor:
I really appreciate the kindness and help with finishing this small change! Thank you for maintaining this project!

PR #1550, opened 2019-10-07
fix: fix uploading photos on iOS 13

### LINKED ISSUE #1546: Cannot attach photos from iOS
On iOS when you use the camera icon to attach photos / take a photo it opens the File explorer and not the photos app or camera. These are files/documents and not the media stored on the device.

nolanlawson:
fixes #1546

This should fix it anyway, but I don't have an iOS 13 device, nor do I have a great setup for debugging iOS (Ubuntu sadface) so I'm mostly just guessing here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,6 +142,7 @@
       "NodeList",
       "NotificationEvent",
       "OffscreenCanvas",
+      "PointerEvent",
       "URL",
       "WebSocket",
       "__assets__",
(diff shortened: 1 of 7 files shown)


iliana:
Reporting that this works as a PWA on iOS 13.1.2 with Pinafore 1.15.0, where it didn’t work in 1.13.1.

PR #1555, opened 2019-10-07
Return a synthetic response for range requests in service workers

### LINKED ISSUE #1547: mp4s don't play in Safari
CC1C7267-1593-4900-BBD9-D4DEA57B9768

I'm away from a real keyboard and can't debug further right now but this behavior is exhibited by Safari if the media is being served from a web server that doesn't support Range Requests. I didn't think Pinafore proxied any content itself, so perhaps there is another explanation.

sgenoud:
Fixes #1547 (hopefully).

So what was going on? The service worker intecepts fetch requests - but for some reason Safari transforms the status returned by the server from 206 to 200.

Capture d’écran 2019-10-07 à 15 58 16

But Safari does not like a 200 for range request. So it decides there is an issue with the video.

To fix it we just follow this [stackoverflow response](https://stackoverflow.com/questions/54138601/cant-access-arraybuffer-on-rangerequest/54207122).

Unfortunately, for some reason I cannot seem to be able to test service worker with the current build setup (there is an error during the SW initialisation that does not seem related to my changes).

### DIFF
--- a/src/service-worker.js
+++ b/src/service-worker.js
@@ -77,6 +77,34 @@ self.addEventListener('activate', event => {
   })())
 })
 
+// from https://stackoverflow.com/questions/54138601/cant-access-arraybuffer-on-rangerequest/54207122
+const returnRangeRequest = request =>
+  fetch(request, { headers: {}, mode: 'cors', credentials: 'omit' })
+    .then(res => {
+      return res.arrayBuffer()
+    })
+    .then(arrayBuffer => {
+      const bytes = /^bytes=(\d+)-(\d+)?$/g.exec(request.headers.get('range'))
+      if (bytes) {
+        const start = Number(bytes[1])
+        const end = Number(bytes[2]) || arrayBuffer.byteLength - 1
+
+        return new self.Response(arrayBuffer.slice(start, end + 1), {
+          status: 206,
+          statusText: 'Partial Content',
+          headers: [
+            ['Content-Range', `bytes ${start}-${end}/${arrayBuffer.byteLength}`]
+          ]
+        })
+      } else {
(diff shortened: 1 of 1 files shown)


nolanlawson:
Awesome, thank you! LGTM

nolanlawson:
Typically for Service Worker I test using the production build rather than the dev build. Certain things just tend to work better in that case.

To test you can run:

    DEBUG=1 yarn build-and-start

Then open up localhost:4002. The `DEBUG` disables minification but keeps everything else.

sgenoud:
Thanks for the tip - it looks like the PR breaks videos, let me fix it before you merge

sgenoud:
Now it works (tested on both Chrome and Safari desktop so far).

nolanlawson:
Awesome, thanks!

sgenoud:
(this is the webkit ticket that tracks this issue: https://bugs.webkit.org/show_bug.cgi?id=186050)

nolanlawson:
Just so I understand: if the backend media host _does_ support range requests, will this PR break that? Range requests are useful for skipping ahead in the media.

sgenoud:
This is not related to the fact that the backend does support range request. In the case I tested, it does. The issue is the service worker layer.

For a reason I do not completely understand, the service worker transforms the 206 of the opaque response from the fetch into a 200. Then, the video element refuses to play (as it needs a 206).

With this patch, when we do not return directly the opaque response from the `fetch` within the service worker. We build the object ourselves and we set explicitely the status code.

Note that this patch does not fix the case where the backend does not support range requests.

nolanlawson:
Thanks for the clarification! Great work. :)

PR #1559, opened 2019-10-08
chore: disable esm cache in travis

nolanlawson:
This seems to add extra time because Travis has to pack a new archive every time ESM changes something due to a file change.


nolanlawson:
Seems useless because I can't seem to disable it or terser webpack plugin's cache either.

PR #1573, opened 2019-10-15
perf: use webpack contenthash for better caching

nolanlawson:
Also lazy-load page-lifecycle because why not.

This needs some testing, but it seems to work okay. It'd work better if we could enable `runtimeChunk: 'single'` but this breaks Sapper, so we might have to work on that later. (It's a bit sad that the main client.js chunk which is around ~70kB has its cache busted every time due to the Webpack runtime chunk.)


nolanlawson:
Eh, let's give this a shot in dev and see what happens

Issue #1590, opened 2019-10-16
"No video with supported format and MIME type found." on videos native interface can play


Archenoth:
The video embedded in this post [here](https://awoo.space/@thingywott/102974088447947047) seems to play perfectly on awoo.space, but I see the following when I attempt to play it in Pinafore:
![No video with supported format and MIME type found.](https://user-images.githubusercontent.com/1385527/66963466-fde12d00-f030-11e9-9710-6f0d463c5b86.png)

The following are the headers that Pinafore sends and receives for this video:
![Request headers for Pinafore's inerface](https://user-images.githubusercontent.com/1385527/66963608-51ec1180-f031-11e9-8368-f25a6a15750a.png)

For comparison, awoo.space's main interface has these:
![Request headers for awoo.space's interface](https://user-images.githubusercontent.com/1385527/66963678-7f38bf80-f031-11e9-92fd-1ea75401a2e3.png)

It also seems to chunk the video:
![A list of requests for the same video, chunked across these requests. They all return 206 except the last one which has no response](https://user-images.githubusercontent.com/1385527/66963704-924b8f80-f031-11e9-8d91-52e4f513618b.png)

(Also, the last request never seems to get a response?? Firefox doesn't show me the headers it sends either)

Any idea what's going on here?

This is on Firefox 69.0.2 (64-bit) on Windows 10 Pro. (10.0.18362 Build 18362)

nolanlawson:
Thanks for the bug report. I cannot reproduce in Firefox 69.0.3 on either Windows 10, Ubuntu Linux, or Android. My Windows version is 1809 (17763).

Note that I am reproducing via toot.cafe, by pasting the URL to the toot into the search bar (because I don't have an awoo.space account).

So the problem could possibly be:

- awoo.space's media host (compared to toot.cafe's media host)
- my Windows version (17763 vs 18362)
- something else

My hunch is that it has something to do with the range headers or the `type=*` on the `<video>` tag which were added to fix Safari bugs. (#1537 #1555).

By any chance, could you try in another browser? Or try using another instance which is connected to `awoo.space`, rather than `awoo.space` itself? That may help narrow down the cause of this bug.

nolanlawson:
I do notice that in your screenshot, a 200 is returned by Pinafore rather than a 206 by awoo.space. So it could indeed be #1555 that is the cause.

nolanlawson:
[...]

PR #1593, opened 2019-10-17
fix: remove type="*" from videos

nolanlawson:
I never had any evidence that this actually fixed the iOS video issue (I think #1555 actually fixed it), and I suspect that this may be causing #1590.


nolanlawson:
This doesn't seem to help #1590, but I just confirmed that iOS 12 works with or without this thing, so it makes me nervous to have it if I'm not sure that it's doing anything. Let's remove it.

PR #1594, opened 2019-10-18
fix: only do range request override in Safari

### LINKED ISSUE #1590: "No video with supported format and MIME type found." on videos native interface can play

nolanlawson:
should fix #1590

#1555 was a targeted fix for Safari, so we should disable it in non-Safari browsers, since it seems to be causing problems.

/cc @sgenoud

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,7 @@
       "NotificationEvent",
       "OffscreenCanvas",
       "PointerEvent",
+      "Response",
       "URL",
       "WebSocket",
       "__assets__",
--- a/src/service-worker.js
+++ b/src/service-worker.js
@@ -13,6 +13,8 @@ const timestamp = process.env.SAPPER_TIMESTAMP
 const ASSETS = `assets_${timestamp}`
 const WEBPACK_ASSETS = `webpack_assets_${timestamp}`
 
+const isSafari = /Safari/.test(navigator.userAgent) && !/Chrom/.test(navigator.userAgent)
+
 // `static` is an array of everything in the `static` directory
 const assets = __assets__
   .map(file => file.startsWith('/') ? file : `/${file}`)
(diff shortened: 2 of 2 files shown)


nolanlawson:
This PR is deployed to https://pinafore-nolan.nolanlawson1.now.sh for temporary testing

nolanlawson:
Also potentially even in Safari we should not completely strip the `Accept` headers like that StackOverflow answer suggests. https://stackoverflow.com/questions/52087208 I'm reading through the WebKit bug and frankly I'm not sure the best way to solve this for Safari.

nolanlawson:
It just occurred to me that if the target media server does not support CORS, then I don't even understand how this fix ever worked. It relies on making a CORS-mode request to the mp4 file, then inspecting the bytes, slicing them, and returning a range. Maybe this is why I can't reproduce - it probably depends heavily on the configuration of the target media server, and all of the media servers I run (e.g. for toot.cafe) have CORS enabled.

nolanlawson:
I have no idea how we can fix the combination of Safari + servers that don't support CORS.

However, this PR is definitely an improvement, because at least for non-Safari browsers, we don't break when the media server doesn't support CORS.

I'll have to open up a new PR to solve the Safari+nocors issue.

Issue #1457, opened 2019-08-30
Reconsider allowing tesseract to store a 23MB blob in IndexedDB


nolanlawson:
tesseract.js stores a 23.4667MB Uint8Array containing the English training data in IndexedDB. This may be a recipe for blowing through our storage budget.

We could consider [estimating quota](https://developer.mozilla.org/en-US/docs/Web/API/StorageEstimate/quota) and/or just not letting it store that data in IndexedDB and using the regular HTTP cache instead.

nolanlawson:
Filed a bug on tesseract: https://github.com/naptha/tesseract.js/issues/334

nolanlawson:
This is supported in latest tesseract, upgrade is in progress: #1596

PR #1596, opened 2019-10-20
chore: update tesseract to the latest version

nolanlawson:
This PR is not ready yet; there seems to be a bug in tesseract that is preventing us from upgrading: https://github.com/naptha/tesseract.js/issues/352


nolanlawson:
So it turns out that https://github.com/naptha/tesseract.js/issues/352 is just because I was using `lolcathost` instead of `localhost`. Hopefully this will actually work when deployed to `dev.pinfaore.social`, but we should probably test first.

nolanlawson:
Tested with `now`, it works! :) https://pinafore-nolan.nolanlawson1.now.sh/

nolanlawson:
this will fix #1457 because we'll store an 11mb gziped file in cache storage instead of a 23mb ungzipped file in IDB.

PR #1603, opened 2019-10-25
Further optimize used SVGs

alrra:
@nolanlawson I didn't wanted to change too much, so I only optimized the used SVGs (removing the unused SVGs, SVGO, and so on is up to you).

---

After the build step, for `__sapper__/export/icons.svg` the stats are as follows: 

```text
size before:        26201 B
size after:         23949 B
---------------------------
size reduction:     2252 B (8.60%)
```

The icons are very small, so there isn't much to reduce.


nolanlawson:
Thank you! Nice savings; I'll take it. :)

PR #1618, opened 2019-11-01
Revert "perf: Further optimize used SVGs (#1603)"

nolanlawson:
This reverts commit f17386c8cb25225cd86f0a3d5564f41273d3653c.

Fixes #1616

PR #1620, opened 2019-11-01
fix: revert circle.svg for now to fix blank svg

### LINKED ISSUE #1616: Filled circle SVG shows as invisible
This seems to have been regressed by #1603 

Screenshot from 2019-10-31 18-53-12

I'm not sure exactly what happened, and reverting the single SVG (`circle.svg`) results in an error during build:

```
TypeError: Cannot read property 'append' of undefined
    at readSvg (/home/nolan/workspace/pinafore/bin/build-svg.js:21:5)
```

nolanlawson:
fixes #1616

follow-up to #1618


nolanlawson:
This just reverts the circle.svg back to the original version minus its width/height which aren't necessary since we have the viewbox.

src/thirdparty/font-awesome-svg-png/white/svg/circle.svg
```diff
@@ -1 +1,2 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1792 1792"><circle cx="50%" cy="50%" r="42.9%" fill="#fff"/></svg>
\ No newline at end of file
+<?xml version="1.0" encoding="utf-8"?>
```
alrra: This line is not needed.

```suggestion
```

src/thirdparty/font-awesome-svg-png/white/svg/circle.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1792 1792"><circle cx="50%" cy="50%" r="42.9%" fill="#fff"/></svg>
\ No newline at end of file
+<svg viewBox="0 0 1792 1792" xmlns="http://www.w3.org/2000/svg"><path d="M1664 896q0 209-103 385.5t-279.5 279.5-385.5 103-385.5-103-279.5-279.5-103-385.5 103-385.5 279.5-279.5 385.5-103 385.5 103 279.5 279.5 103 385.5z" fill="#fff"/></svg>
```
alrra: Maybe the problem is that something expected a `path` and not a `circle`? In any case, the path can be shorter:

```suggestion
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 17920 17920"><path fill="#fff" d="M16640 8960q0 2090-1030 3855t-2795 2795-3855 1030-3855-1030-2795-2795-1030-3855 1030-3855 2795-2795 3855-1030 3855 1030 2795 2795 1030 3855z"/></svg>
```

nolanlawson: Yeah that could be it. I think the build script assumes a path? In any case I saw the bug in both Firefox and Chrome, so I don't think it's a browser issue.

nolanlawson: You're right. OK that is a much simpler fix.

PR #1622, opened 2019-11-02
perf: remove inline script

nolanlawson:
The existence of an inline script was predicated on two ideas:

1. we could avoid the flash of unstyled CSS by quickly loading the theme CSS
2. ditto for when the user has modified the "local" link in the nav button to be something else

But it has some downsides:

1. It's ~3kB, which cannot be bytecode-cached by the browser as easily as normal scripts
2. It adds complexity to the build and dependencies, because I have a separate Rollup build process on top of the Webpack build process plus some CSP logic for the inline script checksum

When removing it, I observe:

- basically no difference in Chrome; Chrome seems to do a good job of not flashing the unstyled CSS
- a very small difference in Firefox; Firefox will typically flash the unstyled CSS when refreshing, but if I remove the inline script then it always flashes

I need to test in Safari and noodle on this a bit more, but I'm just curious if the tests will even pass with this removed.


nolanlawson:
The inline script adds about 1s to the build time (1.1s for building the template.html versus 100ms), and the difference in gzipped size for the index.html is 7581 bytes versus 6314 bytes.

nolanlawson:
Hm, there is kind of a "flash of unstyled scrollbar" in Chrome if you disable the custom scrollbar and refresh.

nolanlawson:
I've looked at the traces, and we're talking about 0.4ms of compile time even on 6x slowdown in Chrome, and plus we've established that the inline script fixes some flashes of unstyled style/scrollbar/etc. So I think actually this change is not a good change and we should keep the inline script.

Issue #1629, opened 2019-11-09
Safari+VoiceOver: autosuggest doesn't announce aria-activedescendant changes


nolanlawson:
As described in [this blog post](https://nolanlawson.com/2019/11/05/what-ive-learned-about-accessibility-in-spas/), the autosuggestion list in the compose box currently doesn't work quite right in Safari with VoiceOver. It does work fine in NVDA with Firefox though.

Background: https://github.com/nolanlawson/pinafore/pull/1513#issue-320087960

I think it might be related to this WebKit bug: https://bugs.webkit.org/show_bug.cgi?id=167671

PR #1630, opened 2019-11-10
fix: improve autosuggest a11y

nolanlawson:
some progress on #1629

- works in Chrome on NVDA now
- works in Chrome on VoiceOver now
- shorter aria-labels, don't repeat information like "1 of 3", because it causes the screen reader to speak too frequently, e.g. when the selected result hasn't changed but the number of results has. Also both NVDA and VoiceOver already speak this information.
- stop doing a fancy fade animation, just show and hide the input instantly. I worry it confuses screen readers to have the aria-hidden attribute in there at all. Also it's not really that nice of an animation.
- stop using a single id to identify the active descendant - give immutable IDs and then update the aria-activedescendant instead. I think this is what fixed Chrome.

### DIFF
--- a/src/routes/_components/compose/ComposeAutosuggest.html
+++ b/src/routes/_components/compose/ComposeAutosuggest.html
@@ -1,6 +1,4 @@
-<div class="compose-autosuggest {shown ? 'shown' : ''} {realm === 'dialog' ? 'is-dialog' : ''}"
-     aria-hidden={!shown}
->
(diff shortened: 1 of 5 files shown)


nolanlawson:
NVDA+Firefox still works the same as before.

However, VoiceOver+Safari is still not working really at all. I've tried lots of different tweaks, but I still can't get it to work. However, VoiceOver does work with Chrome, indicating this may be a Safari issue, not a VoiceOver issue.

Issue #1633, opened 2019-11-10
Clean up use of toggle buttons


nolanlawson:
After long discussions with @marcozehe in [this thread](https://toot.cafe/@nolan/103110863561375513), I think our use of toggle buttons can be improved in this way:

1. The favorite/boost/follow buttons should just be normal buttons that change their label (e.g. "favorite" vs "unfavorite", "follow" vs "unfollow"). The `aria-pressed` attribute is not really necessary if we change the label, and plus this would make the `aria-label` the same as the `title` which is much less confusing from a dev standpoint (less awkward code) and a user standpoint (some screenreaders read both the aria-label and the title).
2. Ditto for the "add content warning" / "add poll" buttons &ndash; we can just change the label to "remove content warning" / "remove poll."
3. Ditto for the reply button, which can be something like "open reply" and "close reply." Additionally, we can use [`aria-expanded`](https://www.w3.org/WAI/GL/wiki/Using_aria-expanded_to_indicate_the_state_of_a_collapsible_element) to indicate that there is some UI that is being expanded or collapsed.
4. The pin timeline / choose instance UIs should be radio buttons rather than toggle buttons, because logically you can only select one at a time.
5. In the media modal, the toggle buttons should be either radio buttons or nav links, because once again you can only select one at a time.

So in general, it seems there is no reason for us to use toggle buttons anywhere, since I've just listed above all of our toggle buttons. :slightly_smiling_face:

MarcoZehe:
This sounds quite complete to me.

nolanlawson:
With #1647 and #1648 I would consider this issue fixed.

The only thing I _didn't_ do was `aria-expanded`, because I looked through [the docs](https://www.w3.org/WAI/GL/wiki/Using_aria-expanded_to_indicate_the_state_of_a_collapsible_element) and wasn't quite sure what it would actually add to the "close reply" / "open reply" button.

I also just made the media buttons a list as it was simpler than `aria-current`, which felt odd to me given that the "go left," "go right," "first item," and "second item" buttons were all mixed together. Instead I just put an explicit label saying which one is the current one, and put all the buttons in a list.

PR #1634, opened 2019-11-10
fix: use radiogroup for instance switcher

nolanlawson:
Progress on #1633

This is just s quick sketch of implementing radio buttons for the instance switcher.

One very tricky thing is that the UI is made up of a list where each item has two interactive elements: a link that goes to the instance page, and a button to switch to that instance. It's only the button that should be a radio button, and it's very weird to have links mixed in there.

My current solution is just to wrap the whole thing in a radio group, and then make it so that if any radio button is focused, then you can press up and down to change the radio checked value, as with a standard radio control. But you cannot focus the radio group itself.

Testing in NVDA and VoiceOver, it seems okay. At least it speaks how many items are in the radio group and which one is checked. But I may need to noodle on this and test it more.

### DIFF
--- a/src/routes/_components/radio/RadioGroup.html
+++ b/src/routes/_components/radio/RadioGroup.html
@@ -35,8 +35,8 @@
       className: ''
     }),
     computed: {
-      ariaOwns: ({ length }) => (
-        times(length, index => `radio-group-button-${index}`).join(' ')
+      ariaOwns: ({ length, id }) => (
+        times(length, index => `radio-group-button-${id}-${index}`).join(' ')
       )
     }
   }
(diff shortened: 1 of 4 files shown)


saper:
I don't use any assistive technologies but this highlighted usability problem I've had with the switcher. 70% of the time I click accidentally on a link intending to switch and forgot about the star button which is smaller. 

I am somewhat confused about the role of the links themselves; for me it would be fine to get rid of them completely and use links with account names to switch.

Maybe only the blurb of the currently active instance should be visible and change whenever the switch occurs.

nolanlawson:
@saper Hmmm if we get rid of the names, then I would have to change the design somehow and probably you'd only be able to edit the instance settings for the instance you're currently on. That might make more sense, but it would be a big redesign whereas this PR is narrowly focused on improving screen reader accessibility.

nolanlawson:
Overall this solution is not my favorite in the world (it's a bit awkward to have the radio buttons mixed in with the links), but it is marginally more semantic than the existing code in that it communicates that these are radio buttons and which one is checked.

PR #1649, opened 2019-11-23
fix: make autosuggest list appear over modal dialog

### LINKED ISSUE #1645: Autosuggest in modal is cropped
Needs to not be cropped, should probably overlap the modal.

Screenshot from 2019-11-23 08-49-17

nolanlawson:
fixes #1645

Could be broken, needs to be tested in more browsers, but actually this helped clean up the CSS styles quite a bit, because logically it is kind of a globally-absolutely-positioned element.


nolanlawson:
Looks good in Firefox and Epiphany, let's merge

nolanlawson:
Looks good in Edge, VoiceOver and NVDA work the same as before

Issue #1652, opened 2019-11-24
Avoid uppercase input for instance url input on iOS


Krinkle:
When logging in from Mobile Safari (latest iOS 13.x), the Shift key is enabled by default for the first character (as iOS does for most textual input when configured in a Latin language; to ease forming of sentences).

<img src="https://user-images.githubusercontent.com/156867/69489914-7d0b2180-0e77-11ea-97ad-42dc30a00db7.PNG" height="250">

I habitually disabled the Shift-key upon seeing it unexpectedly enabled, just in case uppercase would cause a problem. Given the case-insensitive nature of domain names, I suppose letting it enter `Mastodon.technology` would have worked as well, but remains somewhat confusing to see.

I found the code at [_pages/settings/instances/add.htm](https://github.com/nolanlawson/pinafore/blob/95ef639b214298e9479c36ab4335cde8d283af60/src/routes/_pages/settings/instances/add.html#L28) and saw the `inputmode="url"` attribute is already set (per  https://github.com/nolanlawson/pinafore/pull/1184). I would've expected that to be enough, but apparently not.

Perhaps we can try setting `autocapitalize="none"` as well. ([HTML spec](https://html.spec.whatwg.org/multipage/interaction.html#autocapitalization))

nolanlawson:
Thanks for the tip! I'm wondering if we should disable `autocomplete` and `autosuggest` as well? I believe all three are just for iOS Safari, by the way. https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/Attributes.html

Krinkle:
Yeah, `autocorrect` seems like one we could do without here, especially given it can sometimes alter the input in real-time without the user realising. 

The `autocomplete` feature might still be useful, however. iOS often provides me completion suggestions even for e-mail addresses, usernames, and (in this case) domain names. These generally don't apply automatically but rather provide a button to insert a whole "word".

Krinkle:
While `autocorrect=off` appears to be specific to Apple Safari, it was standardised as `spellcheck=false` ([Mozilla bug](https://bugzilla.mozilla.org/show_bug.cgi?id=943690), [HTML spec](https://html.spec.whatwg.org/multipage/interaction.html#attr-spellcheck), [MDN support](https://caniuse.com/#feat=spellcheck-attribute)). I'll amend the patch to set that as well.

PR #1653, opened 2019-11-24
fix: Avoid autocapitalize keyboard for instance name

Krinkle:
On iOS Safari (possibly other browsers as well), auto-capitalization is enabled
even for input fields with inputmode=url.

In my testing, the inputmode attribute itself is supported and working (in so
far that iOS does consider it and provides me a keyboard optimised for URL
inputs, with quick buttons for slash and .com etc).

The autocapitalize is supported on Safari, per:
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/Attributes.html#//apple_ref/doc/uid/TP40008058-autocapitalize

Fixes https://github.com/nolanlawson/pinafore/issues/1652.

### DIFF
--- a/src/routes/_pages/settings/instances/add.html
+++ b/src/routes/_pages/settings/instances/add.html
@@ -25,7 +25,7 @@
       </noscript>
 
       <label for="instanceInput">Instance:</label>
-      <input type="text" inputmode="url" id="instanceInput"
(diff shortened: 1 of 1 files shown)


nolanlawson:
Awesome, thanks!

PR #1664, opened 2019-12-06
fix(service-worker): Make API call to the right domain

### LINKED ISSUE #1663: Rich push notifications not shown for certain instances
```javascript
self.addEventListener('push', event => {
  event.waitUntil((async () => {
    const data = event.data.json()
    const { origin } = new URL(data.icon)

    try {
      const notification = await get(`${origin}/api/v1/notifications/${data.notification_id}`, {
        Authorization: `Bearer ${data.access_token}`
      }, { timeout: 2000 })

      await showRichNotification(data, notification)
    } catch (e) {
      await showSimpleNotification(data)
    }
  })())
})
```

The issue is `const { origin } = new URL(data.icon)` - this used to work fine when the icon was hosted on `toot.cafe`, but now it's hosted on `assets.toot.cafe`, so the request will fail. I think we can use `event.target.origin` instead.

sorin-davidoi:
Closes #1663.

### DIFF
--- a/src/service-worker.js
+++ b/src/service-worker.js
@@ -169,7 +169,7 @@ self.addEventListener('fetch', event => {
 self.addEventListener('push', event => {
   event.waitUntil((async () => {
     const data = event.data.json()
-    const { origin } = new URL(data.icon)
+    const { origin } = event.target
 
     try {
       const notification = await get(`${origin}/api/v1/notifications/${data.notification_id}`, {


nolanlawson:
Awesome, thanks!

nolanlawson:
Reopening based on https://github.com/nolanlawson/pinafore/issues/1663#issuecomment-563359924

Seems like maybe we need to solve the whole "register push notifications for multiple instances at once" so that we can keep track of which instance is pushing which notifications? @sorin-davidoi

PR #1688, opened 2019-12-23
fix: remove remount

nolanlawson:
We don't actually need it; I can mount and unmount the React component myself and avoid stringifying all the props.


nolanlawson:
Perf fix: I need to fix cacheFirstUpdateAfter logic so that it doesn't set the custom emoji twice and thus render twice.

PR #1689, opened 2019-12-24
perf: use emoji-mart fork for now

nolanlawson:
Because I want to remove the implicit core-js dependency and not wait for an emoji-mart update.


nolanlawson:
Related: https://github.com/missive/emoji-mart/pull/394

PR #1690, opened 2019-12-24
fix: remove intrinsicsize in favor of width/height

nolanlawson:
`intrinsicsize` is no longer on the standards track, but intrinsic width/height is: https://chromestatus.com/feature/5695266130755584


nolanlawson:
Tested in Chrome, Firefox, and Epiphany, and seems to work well. It does see to get rid of the layout shifting in Firefox and Chrome.

Issue #278, opened 2018-05-07
Custmizable theme colors 


shellkr:
It would be nice to have ability to chose colors of the theme yourself. It shouldn't be that many. I think it is perhaps four colors or something like that. I had trouble to find a suitable theme among the current available ones. I like dark themes and there wasn't many like that to chose from.

nolanlawson:
It seems possible to set a few CSS variables to achieve this, but I'm not sure if we would need multiple color pickers or what in order to implement it. May be worth investigating though

Hund:
I didn't see this theme when I search for the topic. This would be great! :)

nolanlawson:
Thinking about this more, it would be great if this were possible, but right now themes are authored using SASS, so for full compatibility we would have to run SASS on the client side, which may be prohibitive (large code size, computationally expensive). Perhaps though we could implement some kind of "sass lite" that only supports darken/lighten (which are basically the only two features I use). Interesting idea, not sure how to pull it off!

PR #1694, opened 2020-01-01
theme improvements

petabyteboy:
**fix: ignore hidden files in scss/themes**
  When editing a theme file in vim, a .file.swp file is created and would
  crash the sass compiler. This ignores any hidden files in the directory,
  since they are probably not theme files.
 
  **feat: improve button colors in various themes**
  It was very difficult to differentiate pressed buttons from unpressed
  buttons with the lightness difference. This changes the unpressed
  buttons to a grey tone, while the pressed buttons use a variation of the
  theme color.
 
  **feat: add real pitch black theme** 
  On AMOLED displays the pitch black theme was unsatisfying, I have
  renamed it to "Ocean Blue" and added a new theme that uses #000 as main
  theme color.

### DIFF
--- a/bin/build-sass.js
+++ b/bin/build-sass.js
@@ -28,7 +28,7 @@ async function compileGlobalSass () {
 }
 
 async function compileThemesSass () {
-  const files = (await readdir(themesScssDir)).filter(file => !path.basename(file).startsWith('_'))
+  const files = (await readdir(themesScssDir)).filter(file => !path.basename(file).startsWith('_') && !path.basename(file).startsWith('.'))
   await Promise.all(files.map(async file => {
     let css = await renderCss(path.join(themesScssDir, file))
     css = cssDedoupe(new TextDecoder('utf-8').decode(css)) // remove duplicate custom properties
--- a/src/routes/_static/themes.js
+++ b/src/routes/_static/themes.js
@@ -90,6 +90,12 @@ const themes = [
     color: '#282C37'
   },
   {
+    name: 'oceanblue',
+    label: 'Ocean Blue',
+    dark: true,
+    color: '#03131b'
+  },
+  {
     name: 'pitchblack',
     label: 'Pitch Black',
     dark: true,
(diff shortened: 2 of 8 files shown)


nolanlawson:
Thanks for the PR! A few thoughts on this:

1. Skipping `.swp` files is a great change, no reason we should not do this.

2. For the pressed vs unpressed state I agree that for some themes it's not easy to distinguish the two. The approach I've typically taken is a onesy-twosy fix for each theme to try to increase contrast, but arguably I could go further for certain themes. In this PR I would prefer not to make the default button color a flat gray because to me it looks a little off (looks like a "disabled" state to me), but I appreciate the effort

Screenshot from 2020-01-12 09-47-19

3. Pitch Black was explicitly designed for a certain type of screen (mobile OLED, details here: https://github.com/nolanlawson/pinafore/pull/727), not necessarily to literally be `#000`. I don't own such a screen so can't confirm whether or not this PR is an improvement for those types of screens.

4. Ocean Blue looks neat, but maybe the default button state doesn't have high enough contrast against the background? Seems a little hard to read, would fail WebAIM guidelines: https://webaim.org/resources/contrastchecker/?fcolor=17191F&bcolor=000000

Screenshot from 2020-01-12 09-50-48

I think possibly this PR should be broken up into separate commits. Also I wonder if Ocean Blue and your custom Pitch Black wouldn't make more sense as "custom" themes that you can apply to your own instance of Pinafore? At some point I would like to cap the number of built-in themes, and I worry that the list is already getting a bit long. #278 would be a big help here.

petabyteboy:
Thanks for your review. I will close this for now, since the main purpose of the PR (changing the button color) will not be accepted. I will probably end up hosting my own instance with this change, because it makes pinafore much easier to use for me.

You say the pitch black theme was added for OLED screens. I use multiple devices with OLED screens, and the changes I made to the theme were specifically to fix this theme on OLED screens.
Since you seem to be unfamiliar with the way black pixels work on OLED screens: On those screens, pixels that are fully black (#000) do not emit light at all. There is no background light and every pixel can be enabled or disabled individually, so they become indistinguishable from the screen border, and also stop consuming energy. The pitch black theme that is currently implemented in pinafore does not use #000 as background, so the pixels light up, the border of the screen is clearly visible and the pixels consume energy.

PR #1702, opened 2020-01-29
feat: change pitch black theme

petabyteboy:
On AMOLED displays the pitch black theme was unsatisfying. I have
changed the theme to use #000 as main theme color, which completely
stops the OLED pixels from emitting light, thus saving energy and
on some devices making the background indistinguishable from the screen
border.


petabyteboy:
In this version, I just changed the existing theme, without creating a copy of it.

nolanlawson:
Looks good, thank you!

PR #1743, opened 2020-04-25
feat: aria-labels and buttons contain more media info

### LINKED ISSUE #1733: [Accessibility] Include information about images in long label
I'm using pinafore almost exclusively with my screen readers browse mode turned off with the help of the excellent keyboard shortcut support.

One problem I'm running into is that the long label doesn't include any information about media attached to a toot, even if it's described. Sometimes, it's obvious that a toot contains an image at which point I'll switch to browse mode to check whether it has a description, but this isn't always the case.

So, I think the long label for a toot should include information about any attached media. This could come directly after the toot text. A good idea for the format could be something like:
Image/Gif/Video (if there's more than 1 attachment, also number them), a colon so that speech would pause, and if the media has a description, include it next, otherwise say no description.

Knowing if something is a static image or a gif is useful because even if a static image has no description, a blind person would know that they can try to run additional tools on it, like OCR, which might still give a useful result, which isn't anywhere nearly as likely if it's a GIF.

nolanlawson:
fixes #1733

I implemented this slightly differently from @pitermach's suggestion, but I think in general it's an improvement.

- long aria labels now contain "has media" and "has poll" to indicate whether there's media or a poll attached
- buttons to open the media modals now specify the type. it says "play" for video/audio and "show" for images/gifs, and then for gifs it says "show animated image"

The main reason I didn't put all this information into the long aria labels is that there is complexity in whether images are shown or not due to spoilers or settings, and this state is currently stored deep in the component hierarchy where it's difficult to access. In the future this could certainly be improved, but at least for now the information is better surfaced.


nolanlawson:
Travis integration not working for some reason, but the build passed

PR #1746, opened 2020-04-26
fix: fix nav links in Voice Control on iOS

### LINKED ISSUE #1735: Tab buttons at top aren’t clickable with iOS Voice Control on an iPhone
Filing this issue on behalf of a friend, who can’t access the tabs at the top since they don’t register as a tappable/clickable area.

This might be happening because the tabs don’t have text displayed due to screen width. The tappable areas are detected on an iPad.

Post with video and context is at https://mst3k.interlinked.me/@artemis/103933244302642043

> When I say "show numbers" all tappable things on screen should get a number and I can say a number to tap it. but the bug is the nav bar items don't get numbers. Same for "show names" but names come from aria labels or tries to guess if they're missing
>
> The url is dev but it happens on dev and not dev

nolanlawson:
fixes #1735

After playing around with Voice Control on iOS 13, it seems that Voice Control really doesn't understand aria-label at all, but it understands text inside of the link just fine. So we can exploit that to just move the text off-screen instead of using `display: none`.

Incidentally, Voice Control also doesn't seem to like small tap targets very much, so it still won't necessarily show the names or numbers when you say "show names" or "show numbers" for all links. But I also tested on a very small phone (iPhone SE first gen), so presumably on a larger phone this would work better. And in any case, this PR at least makes some of the links work correctly, even on a small screen.

Long-term, I assume this is just a bug that Apple will fix. aria-label should be exactly the same as hidden text.

### DIFF
--- a/src/routes/_components/NavItem.html
+++ b/src/routes/_components/NavItem.html
@@ -92,7 +92,18 @@
 
   @media (max-width: 991px) {
     .main-nav-link .nav-link-label {
-      display: none;
+      /* Copied from the sr-only styles in global.scss
+       * the reason we explicitly leave this <span> in is because Voice Control on iOS does not
+       * understand aria-labels very well, but it understands hidden text just fine
+       */
+      position: absolute;
+      width: 1px;
+      height: 1px;
+      padding: 0;
+      margin: -1px;
+      overflow: hidden;
+      clip: rect(0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


faithanalog:
Thanks! I agree it's very likely just an iOS bug  - Voice Control is actually new as of iOS 13 and its got quite a few rough edges to even out still

PR #1747, opened 2020-04-26
perf: make timeline rendering less janky

nolanlawson:
1. Ensures all statuses are rendered from top to bottom (no more shuffling-card-effect rendering)
2. Wraps all individual status renders in their own requestIdleCallback to improve input responsiveness especially only slow devices like Nexus 5.

I'm still mulling over 2 but 1 is a solid improvement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,6 +163,7 @@
       "matchMedia",
       "performance",
       "postMessage",
+      "queueMicrotask",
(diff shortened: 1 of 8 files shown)


nolanlawson:
Backed down from 2, made it mobile-only due to the scrollbar resizing in a weird way on desktop.

PR #1749, opened 2020-04-27
fix: fix display of <br>s for Pixelfed

nolanlawson:
Turns out Pixelfed likes to inject tons of `<br>`s, but we can just hide them.


nolanlawson:
Hm, this actually breaks a test, looks like I will need to get creative

nolanlawson:
I really can't figure out why it looks so different on Pinafore vs Mastodon. But it's really not a big deal

PR #1757, opened 2020-05-02
update sapper to fix SNYK-JS-SAPPER-561051

nolanlawson:
This vuln doesn't really affect us sync we use `sapper export` to build static files (rather than actually running Sapper on the server side), but we may as well fix it.

Sadly had to fix via cherry-pick because latest Sapper is incompatible with Svelte v2. https://github.com/nolanlawson/sapper/commit/99798006c3a00b9f95ede212f844c75ee7cf1bb0

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "rollup": "^2.0.0",
     "rollup-plugin-babel": "^4.3.3",
     "rollup-plugin-terser": "^5.2.0",
-    "sapper": "nolanlawson/sapper#for-pinafore-14",
+    "sapper": "nolanlawson/sapper#for-pinafore-20",
     "stringz": "^2.1.0",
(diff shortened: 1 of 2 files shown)


nolanlawson:
Yep I can confirm we're vulnerable in dev mode, but not prod mode.

Run `yarn dev` and then

```
curl 'localhost:4002/client/%2e%2e/%2e%2e/%2e%2e/package.json'
```

It serves up the package.json when it shouldn't. In prod mode we use `express` instead of `sapper` so the vuln doesn't exist.

nolanlawson:
vuln: https://snyk.io/vuln/SNYK-JS-SAPPER-561051

nolanlawson:
Confirmed that the `curl` command no longer works in dev mode

PR #1759, opened 2020-05-02
fix: beef up CSP headers

nolanlawson:
Improves the CSP a bit, following advice from Mozilla Observatory.

- `default-src` is `'none'`
- restrict `form-action` and `frame-ancestors`
- restrict `img-src` and `media-src` to just what we need

Pinafore doesn't need to run in an iframe, and it does need to send XHRs and img/audio/video requests to arbitrary origins, but most stuff we can just block

### DIFF
--- a/bin/build-now-json.js
+++ b/bin/build-now-json.js
@@ -69,7 +69,7 @@ const HTML_HEADERS = {
     "style-src 'self' 'unsafe-inline'",
     "img-src 'self' * data: blob:",
     "media-src 'self' *",
-    "connect-src 'self' *",
+    "connect-src 'self' * data:",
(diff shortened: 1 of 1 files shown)


nolanlawson:
let's add `base-uri 'none'` while we're at it.

nolanlawson:
Ah great, this actually breaks SVGs in Firefox because of a longstanding bug:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1303364
- https://github.com/w3c/webappsec-csp/issues/199

There is no solution apparently except not to use `default-src 'none'`, which Mozilla Observatory explicitly recommends :man_facepalming:

PR #1767, opened 2020-05-06
Docker compose

Shadow8t4:
Added a new file `docker-compose.yml` which defines a docker-compose build for the image defined in the `Dockerfile`. This allows production users to use a single command to build and start an image for Pinafore that exposes a defined port, rather than needing to specify an open port in a docker command.

Additionally, added to the `README.md` an instruction on how to build/start the container.


nolanlawson:
Thanks a lot! :)

PR #1769, opened 2020-05-14
perf: lazy-load the thread context

### LINKED ISSUE #898: Lazy load thread context
Current behavior when opening a thread:

1. Request status and all ancestors/descendants
2. Render entire thread
3. Scroll to focused status

Better approach:

1. Request status and ancestors/descendants
2. Render status as soon as it returns (fast)
3. Lazy-render ancestors/descendants (slow)
4. Scroll if necessary

This would make opening statuses faster. Mastodon currently does this.

nolanlawson:
fixes #898

Fixes something I really should have fixed a long time ago, but I put off because... I dunno.

When a status is not a reply, if you click on it then it should be able to load instantly and let the rest of the thread load lazily. You can tell that the original Mastodon FE already does this.

The only thing I dislike a bit about their implementation is that they do this for replies as well, and I'd prefer for the timeline not to render twice and jump (looks jerky). So we can use the old behavior for that case and block on the network.

This makes tapping most statuses much faster since most are not replies and most are complete singletons anyway (or at least, a lot are).


src/routes/_actions/timeline.js
```diff
@@ -49,7 +104,7 @@ async function fetchTimelineItems (instanceName, accessToken, timelineName, last
     try {
       console.log('fetchTimelineItemsFromNetwork')
       items = await fetchTimelineItemsFromNetwork(instanceName, accessToken, timelineName, lastTimelineItemId)
-      /* no await */ storeFreshTimelineItemsInDatabase(instanceName, timelineName, items)
+      await storeFreshTimelineItemsInDatabase(instanceName, timelineName, items)
```
nolanlawson, line 107: I was trying to be clever by avoiding this `await` and relying on the sync db cache to be updated first, but this is really fragile and it's not a big deal to wait on IndexedDB after we finished waiting on the network

src/routes/_components/list/ListItem.html
```diff
@@ -1,4 +1,17 @@
...
+    transition: opacity 0.2s linear;
+    contain: content; /* see https://www.w3.org/TR/2018/CR-css-contain-1-20181108/#valdef-contain-content */
+  }
+
+  .list-item.shown {
+    opacity: 1;
+    pointer-events: auto;
+  }
+</style>
+<div class="list-item {shown && 'shown'}"
```
nolanlawson, line 14: Forgot to make the non-virtual list have a nice fade-in animation, fixed that while I was here

src/routes/_database/databaseApis.js
```diff
@@ -6,6 +6,6 @@ export * from './timelines/pagination'
 export * from './timelines/getStatusOrNotification'
 export * from './timelines/updateStatus'
 export * from './timelines/deletion'
-export * from './timelines/insertion'
+export { insertTimelineItems, insertStatus } from './timelines/insertion'
```
nolanlawson, line 9: this thing was exporting too many APIs, accidental mistake although not a big deal

nolanlawson:
hm ok I remember now why this was hard - because the thread order is really hard to reproduce client-side for complicated reply chains. not impossible though

PR #1775, opened 2020-05-16
feat: use :focus-visible, add setting to enable/disable it

nolanlawson:
Uses the [`:focus-visible` polyfill](https://github.com/WICG/focus-visible) to disable the focus ring when using mouse/touch interaction as opposed to keyboard interaction. This follows a pretty common pattern across webapps, where the focus ring is only shown when the user is detected to be a keyboard user.

The standard is currently not implemented in any browser, except Chrome with "experimental web platform features" enabled. But if it's never implemented, we can just keep using the polyfill.

I also added a setting to always show the focus ring, disabled by default. I disabled it by default because my hunch is that 1) the average mouse user does not know what a focus ring is and can't be bothered to go into their settings to change things, whereas 2) keyboard users are more likely to be a bit savvier (either because they are power users or because they have learned about how to toggle accessibility settings on sites to behave the way they like).

Similar to reduced motion, I think it's okay to have it disabled by default. Unlike reduced motion, though, I really wish there was some kind of CSS media query to detect this, ala the settings on macOS where you can tell it whether you want to always show the focus ring for all elements (not just inputs).

The situation with focus rings on the web is far from ideal, but I think this PR is a good compromise. As a mouse user, I am just really annoyed by having to always click away to remove the focus ring, and I prefer the more "app-like" experience. But I don't want hotkeys or pressing Esc on modals to behave poorly. This PR gives us the best of both worlds.

### DIFF
--- a/src/routes/_pages/settings/general.html
+++ b/src/routes/_pages/settings/general.html
@@ -4,12 +4,12 @@
   <h2>Media</h2>
   <form class="ui-settings">
     <label class="setting-group">
-      <input type="checkbox"
+      <input type="checkbox" id="choice-never-mark-media-sensitive"
              bind:checked="$neverMarkMediaAsSensitive" on:change="onChange(event)">
       Show sensitive media by default
     </label>
     <label class="setting-group">
-      <input type="checkbox"
+      <input type="checkbox" id="choice-use-blurhash"
               bind:checked="$ignoreBlurhash" on:change="onChange(event)">
       Show a plain gray color for sensitive media
     </label>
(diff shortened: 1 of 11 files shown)


src/routes/_pages/settings/general.html
```diff
@@ -4,12 +4,12 @@ <h1>General Settings</h1>
   <h2>Media</h2>
   <form class="ui-settings">
     <label class="setting-group">
-      <input type="checkbox" id="choice-never-mark-media-sensitive"
+      <input type="checkbox"
```
nolanlawson, line 7: Turns out these IDs were not required

nolanlawson: Oops I'm wrong, these are needed for tests

nolanlawson:
I tested this manually pretty thoroughly, both with and without the polyfill (i.e. using experimental web features flag in Chrome). I'm satisfied with how it works - the polyfill behaves exactly the same as the native implementation, and it matches my expectations regarding Tab/Esc/left-right focus change/up-down hotkeys, etc.

nolanlawson:
Tested `-moz-focusring`. It doesn't seem t consistently work for all elements, and has some ugly artifacts (click a nav button, the focus ring momentarily shows, then hides... wtf?). I'd prefer to just polyfill for Firefox.

PR #1799, opened 2020-06-11
Fix /api/v1/instance requiring authentication in whitelist mode

### LINKED ISSUE #1795: Cannot sign in while instance is in whitelist mode
Whitelist mode is another feature in Mastodon which allows an instance to cut itself off to just a few instances to prevent abuse. Pinafore doesn't login in this mode. I think this is a bug.

koyuawsmbrtn:
Hackily fixes #1795

### DIFF
--- a/src/routes/_actions/addInstance.js
+++ b/src/routes/_actions/addInstance.js
@@ -1,5 +1,5 @@
 import { getAccessTokenFromAuthCode, registerApplication, generateAuthLink } from '../_api/oauth'
-import { getInstanceInfo } from '../_api/instance'
+import { getOnlineMode } from '../_api/instance'
 import { goto } from '../../../__sapper__/client'
 import { DEFAULT_THEME, switchToTheme } from '../_utils/themeEngine'
 import { store } from '../_store/store'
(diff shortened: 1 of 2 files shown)


rgx1:
Thank you **Koyu**! Wonderful.

nolanlawson:
Thanks for opening the issue and the PR, but I'm not sure what the issue is or how this is solving it? It looks like this PR is just hard-coding the instance URL to `mastodon.social`? How does that fix the whitelisting bug?

koyuawsmbrtn:
> 
> 
> Thanks for opening the issue and the PR, but I'm not sure what the issue is or how this is solving it? It looks like this PR is just hard-coding the instance URL to `mastodon.social`? How does that fix the whitelisting bug?

I didn't know how to remove the `/api/v1/instance` check in general and I don't know why it's there in the first place. Removing it in general would fix the issue since even then Pinafore performs some checks against the input. Maybe someone comes up with a better idea, but completely removing it would be nice. I don't know what `get(url, null, { timeout: DEFAULT_TIMEOUT })` produces to make Pinafore happy and just try to authenticate anyway.

nolanlawson:
@koyuawsmbrtn I see, so the whitelisting feature is that the Mastodon server will block API access _until_ you log in? So we cannot assume public access to `api/v1/instance`?

In that case I will need to re-architect Pinafore a bit, but this should be solveable. Do you have an example of a Mastodon instance with this whitelisting feature?

koyuawsmbrtn:
> 
> 
> @koyuawsmbrtn I see, so the whitelisting feature is that the Mastodon server will block API access _until_ you log in? So we cannot assume public access to `api/v1/instance`?
> 
> In that case I will need to re-architect Pinafore a bit, but this should be solveable. Do you have an example of a Mastodon instance with this whitelisting feature?

Exactly. An example instance would be my own, koyu.space.

PR #1801, opened 2020-06-13
Fix favorites

### LINKED ISSUE #850: Favorites are not paginating correctly
The favorites timeline is not paginating past default 20-post limit. This is because the id of the last status that was retrieved is being used as the max_id in the request, but it needs to use the max_id provided in the `Link` header per the [Mastodon API](https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#favourites):

> Note: max_id and since_id for next and previous pages are provided in the Link header. It is not possible to use the id of the returned objects to construct your own URLs, because the results are sorted by an internal key.

As an additional note, other routes require using the `Link` header instead of returned objects. It might be worth testing other routes to be sure they work correctly.

charlag:
fix #850
                                   
This commit fixes invalid assumption that all timelines are sorted by status id.
Some, like favorites or bookmarks are sorted by private server id. To correctly
paginate we must use the Link header.

To work around the issue, offline for favorites was effectively disabled.
Statuses are still inserted into the database but we can't reproduce correct
timeline order.

Some notes about the code: I wasn't sure how to organize new loading behavior so I just did what seemed like the easiest way to do it. Feel free to change/ask to change anything.

I didn't run integration tests yet. I have dev masto environment but it's in the container together with postgres and I think Pinafore wants to extract things from pg so I'm not sure it would work for me.

EDIT: i installed new package with npm and not yarn, not sure if it affects anything
EDIT2: This is the same package for parsing link header as what Mastodon is using.


nolanlawson:
Thanks for opening the PR! Sorry I haven't had a chance to look at this yet; I've been a bit busy recently. I'm sure we can figure out what is causing the test failures and fix them. Thanks a whole bunch for putting in the work to fix this!!

charlag:
No pressure, I work on Tusky for now. I would appreciate a lot if you looked at it when you have time but I wouldn't want you to stress over it

nolanlawson:
Sorry, was a bit caught up with the emoji picker project over the past few weeks! :sweat_smile: I just got a chance to look this PR over, and I'm shocked at how close it is to what I imagined I would have written myself. :slightly_smiling_face: You solved it basically the exact same way I was thinking of solving it.

I'm not sure why the test is broken, but I'll look into it and merge soon; this looks great! Definitely the easiest way to fix favorites is to just make it online-only, and I also like the way the nextId is stored in the store.

nolanlawson:
Looks like `ajax.js` needed to be tweaked; I also went ahead and wrote a test to ensure that favorites pagination works. It worked locally! Nice work. :slightly_smiling_face: 

There still seems like there might be a test failing; I will take a look tomorrow. Thanks again for writing this!

charlag:
Thanks for looking at it, I'm really glad it worked! After that adding bookmarks should be a breeze, even for me.

nolanlawson:
Some more minor fixes were needed, tests are passing now.

Right now I'm just looking into the http-link-header dependency, it pulls in the whole Node `Buffer` dependency so seems a bit large

charlag:
@nolanlawson thanks a lot for fixing it, I'm so lost without types.
About that header dep: ouch. Maybe we should let Mastodon know about this too. The library probably doesn't even have to depend on it.

nolanlawson:
Thanks @charlag! Unfortunately even the `li` library I found is not as small as I would like (it doesn't allow for tree-shaking), but it's still the smallest one I could find. Maybe I will open a pull request on it to add tree-shaking.

Thanks a lot for your contribution; I think we are well on the way to getting favorites and bookmarks to work as well! I'll merge the other PR since I squashed the commits in there.

PR #1804, opened 2020-06-28
feat: use emoji-picker-element, add emoji autocompletions/tooltips

nolanlawson:
Uses `emoji-picker-element` (an emoji picker I just released) to replace `emoji-mart`. Because `emoji-picker-element` exposes the Database as a separate import, we can also re-use it to show native emoji autocompletions and tooltips (`title` attribute).

Memory usage is lower, bundle size is lower, runtime performance is faster. I measured the new picker versus the old picker on a variety of metrics and it seems like an improvement.

Also we can get rid of the `preact` dependency and the Babel transform we were just using for `emoji-mart`.


nolanlawson:
This PR still needs some work, e.g. hiding unsupported native emoji in the autocompletion dropdown.

nolanlawson:
Hmmmm, need to figure out how to install fonts-noto-color-emoji in Ubuntu xenial. I may have to upgrade to bionic.

Ah, the joys of native emoji :sweat_smile:

nolanlawson:
Ack, postgres is broken in Travis on Bionic. I'll have to do some debugging for this.

nolanlawson:
I should put some screenshots here which I can use in the release notes later

custom emoji and regular emoj in one autosuggest:
Screenshot from 2020-06-29 10-17-13

new picker:
Screenshot from 2020-06-29 10-16-08

tooltips for native emoji (desktop only)
Screenshot from 2020-06-29 10-15-38

nolanlawson:
Memory usage before and after (7.7MB vs 3.2MB): 
Screenshot from 2020-06-29 10-24-51

Load time improvement (second load) - 3.9s versus 700ms on 6x CPU throttling

Screenshot from 2020-06-29 10-27-56

nolanlawson:
Small bug - maybe the scrollbar should be dark when in dark mode? Haven't done that for modals historically

PR #1823, opened 2020-07-05
chore: update now to v19

nolanlawson:
Should probably switch to `vercel` soon too


nolanlawson:
might as well update to `vercel`

PR #1837, opened 2020-08-02
autofocus search input field

shinenelson:
it is annoying to have to use `Tab` to focus the search input field on
the search page ( in keyboard mode ).

this changeset adds the `autofocus` tag on the input field so that it is
focused when the search page is brought to view.

I am aware that it is not recommended to use `autofocus` on input fields
for accesibility reasons. However, I would respectfully argue that for a
page which is descriptive and obvious where there is nothing but a single
input field and a button on that page that it does not warrant to follow
the recommendation blindly.

if the accessiblity recommendation has more priority or a 100%
accessibility score is more important, then this pull request may be
declined and closed.

### DIFF
--- a/src/routes/_components/search/Search.html
+++ b/src/routes/_components/search/Search.html
@@ -5,6 +5,7 @@
            placeholder="Search"
            aria-label="Search input"
            required
+           autofocus
            bind:value="$queryInSearch">
   </div>
   <button type="submit" class="primary search-button" aria-label="Search" disabled={$searchLoading}>


nolanlawson:
Thanks for the PR! Looking into this, this actually causes an accessibility warning:

```
Module Warning (from ./node_modules/svelte-loader/index.js):
A11y: Avoid using autofocus (9:11)
 6:            aria-label="Search input"
 7:            required
 8:            autofocus
               ^
 9:            bind:value="$queryInSearch">
10:   </div>
```

The source of the warning is [this](https://www.a11yproject.com/checklist/#avoid-using-the-lesscodegreaterautofocusless/codegreater-attribute).

Additionally, this breaks some tests because, when pressing left or right to change the current column, the focus changes to the search box when reaching the search page, which would cause the left and right buttons to no longer work.

I don't think autofocusing the search box is a bad idea in all cases (some nuance discussed [here](https://www.brucelawson.co.uk/2009/the-accessibility-of-html-5-autofocus/)), but given that we would have to 1) figure out how to disable the Svelte warning, and 2) somehow solve the left/right navigation issue, I am inclined not to merge this PR. The second issue in particular is very tricky, because to fully fix it we would have to somehow detect what input method the user used to arrive at the search page before deciding whether to focus or not (e.g. keyboard versus mouse).

nolanlawson:
An alternative is to combine this with the `/` shortcut (#1838) - maybe when pressing that hotkey in particular, it should change to the Search page _and_ autofocus? In that case, though, this would be kind of a "hidden" feature, so I'm not sure how valuable it would be.

shinenelson:
I couldn't miss the accessibility warning. It was almost in my face every time I ran a build. I had addressed it in the description itself - as mentioned in the article you linked, the warning doesn't make sense when the search input is the only thing on the page.

> 1) figure out how to disable the Svelte warning

quick stack overflow search yielded [this result](https://stackoverflow.com/a/61277660) that does exactly what is required. however, I'm not sure we'd want to do that on a project level, rather on a module level. I don't know if that is possible with this solution though.
if this is a feasible solution ( note : we do not have [rollup-plugin-svelte](https://github.com/sveltejs/rollup-plugin-svelte) as a dependency yet ), I can add the relevant bits of code to this pull request.

> 2) somehow solve the left/right navigation issue
> to fully fix it we would have to somehow detect what input method the user used to arrive at the search page before deciding whether to focus or not (e.g. keyboard versus mouse)

I am a keyboard-driven user. For me, hitting multiple Tabs to get focus to the search input is annoying. However, for a mouse user, it is 2 clicks. I don't think a screen-reader would have too much troubling reaching the search input either. I guess I was being a little selfish on behalf of keyboard users ( sorry about that ).

> An alternative is to combine this with the `/` shortcut (#1838) - maybe when pressing that hotkey in particular, it should change to the Search page and autofocus? In that case, though, this would be kind of a "hidden" feature, so I'm not sure how valuable it would be.

When you say "_that hotkey in particular_", do you mean only the `/` key? Or do you generally mean the hotkey for the **Search** column ( which includes `S` as well )? As I've agreed earlier, I would vote for the feature for keyboard-driven users which means both `/` and `S` but not mouse-clicks. I don't know if the application already has support for detecting the input method being used. I could attempt a fix if pointed in the right direction.

nolanlawson:
Thanks for your thoughtful reply. Here's my proposal:

1. `/` and `s` both navigate to the Search column (page), as well as focus the search input.
2. If you're already on the Search page, then those hotkeys focus the search field
3. Instead of using `autofocus`, let's do a programmatic `element.focus()` call. This will eliminate the Svelte warning, and plus we will probably have to do it anyway to accomplish # 1 and # 2.

As for the implementation details, I admit I'm not sure how best to go about this. The Search page will need some knowledge of _how_ it was navigated to. Probably something as simple as `/` and `s` writing to the store, and the Search page clearing that value from the store on its `oncreate`, would suffice.

Does this work for your use case?

shinenelson:
Yes, it makes sense. As long as I can navigate to the Search column using the search hotkeys ( either of `/` or `s` ), and the search input is auto-focused when navigating to the column, I'm fine.

The current proposal seem to address that well for me.

But like you said, the implementation details are kind of complicated for me to comprehend.

nolanlawson:
The trickiest part is just writing some data to the store. For instance, when the keys are pressed, it would be e.g.:

```js
store.set({userJustPressedSearchHotkey: true})
```

And then inside of the `oncreate` for the search page (`search.html`), it would be e.g.:

```js
if (store.get().userJustPressedSearchHotkey) {
  this.refs.autocomplete.focus() // get a reference to the element using refs and focus it
  store.set({userJustPressedSearchHotkey: false}) // cleanup
}
```

nolanlawson:
Closing in favor of #1855, I implemented the suggestion above ^

shinenelson:
This is great. Thank you for implementing this so quickly. Though I wanted to try to implement it myself, I couldn't find the time to do it myself.

I did test it and it works as expected 👍

PR #1838, opened 2020-08-02
add `/` as a navigation shortcut for search

shinenelson:
`/` is a well-known vi/vim key-binding for search. It is supported by
Firefox for a 'quick find' feature in addition to the main find feature
available with the Ctrl+F key combination. [DuckDuckGo](https://duckduckgo.com) also supports the
key to focus the search bar as well. Other websites like GMail also
focuses the search bar on pressing the `/`. It is not only DuckDuckGo or
GMail alone that supports the key. I've noticed it on multiple websites
but I can only think of these 2 that are popular enough to argue support
over the top of my head. I cannot say if it is a good practice per se;
but I feel that it is a good feature to support.

### DIFF
--- a/src/routes/_components/NavShortcuts.html
+++ b/src/routes/_components/NavShortcuts.html
@@ -5,7 +5,7 @@
 <Shortcut key="g n" on:pressed="goto('/notifications')"/>
 <Shortcut key="g c" on:pressed="goto('/community')"/>
 <Shortcut key="g d" on:pressed="goto('/direct')"/>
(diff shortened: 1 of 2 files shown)


nolanlawson:
LGTM, the only thing it needs is for `/` to be added to the help dialog, which I've done in a separate PR ^

shinenelson:
Whoops, I'm sorry I missed to add the new key binding to the Help dialog. Thank you for catching and adding it.

PR #1841, opened 2020-08-21
[theme] twitter dark

shinenelson:
this is an attempt to theme Pinafore to use the colors that Twitter uses
for their dark mode.
Normal use-cases have been tested manually. I don't think I can possibly
cover all edge-cases.

### DIFF
--- a/src/routes/_static/themes.js
+++ b/src/routes/_static/themes.js
@@ -100,6 +100,12 @@ const themes = [
     label: 'Dark Grayscale',
     dark: true,
     color: '#666'
+  },
+  {
+    name: 'dark-twitter',
+    label: 'Dark Twitter',
+    dark: true,
+    color: '#1B2836'
   }
 ]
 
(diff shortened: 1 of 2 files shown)


nolanlawson:
Thank you very much for this PR and the other Twitter theme (#1842). In general, I think both themes look great. However:

1. We already have a lot of themes.
2. In the past, I've been welcoming of new themes, but recently I've been more and more reluctant to add new ones, since they add a performance overhead (new CSS files are eagerly cached in the Service Worker) and cognitive overhead (more UI on the screen).
3. Since Mastodon is a Twitter alternative, I'm a bit uncomfortable to make explicit reference to Twitter in the app.

So I wonder if these themes wouldn't make more sense as customizations in your own fork of Pinafore? In the long term, this may also be covered by custom user themes (https://github.com/nolanlawson/pinafore/issues/278).

Again, thanks for submitting the PRs!

shinenelson:
Sure, no problem. I understand your position.

Psst, I have a little bit more ambitious goals for my fork of pinafore and this was a hint at it ( clue : think halcyon )

PR #1851, opened 2020-08-29
fix(VirtualList): fix some TODOs

nolanlawson:
Just some housekeeping. I didn't really understand how Svelte worked at the time I wrote this code.


nolanlawson:
Just realized this adds a memory leak. Need to add a test to catch that particular leak.

PR #1859, opened 2020-09-01
fix: pinning the bookmarks page

### LINKED ISSUE #1858: Pinning Bookmarks pins local timeline instead
Pinafore 1.19.0 using iOS 13.5.1.

Can consistently reproduce after restart.

arethsu:
Pinning the bookmarks page would accidentally pin the local timeline page due a forgotten `else if`. :heart:

Fixes #1858


nolanlawson:
Oops! Thanks very much. :+1:

PR #1860, opened 2020-09-02
fix: tweak maskable PWA icons

nolanlawson:
I feel like the maskable icons look a bit better if they're a bit smaller.


nolanlawson:
Some before-and-after screenshots (before above, after below)

out

Screenshot_20200901-224156_Trebuchet

PR #1867, opened 2020-09-06
fix: remove LEGACY mode

nolanlawson:
I originally implemented legacy mode to support KaiOS. Since I never got around to actually building a KaiOS app, and since this is starting to seem like a distant possibility, and since by the time I get it done maybe KaiOS will actually upgrade to a modern version of Gecko, these dependencies are starting to feel like more trouble than they're worth. This removes 11 dependencies.


nolanlawson:
Ugh, but then again maybe I should just build the danged KaiOS app.

nolanlawson:
Yeah KaiOS support was a beautiful dream, but it's just so hard to get it working properly because it's such an old browser engine. Trying to fix all the places where the layout/style is messed up, and it just feels endless: b15ce117

nolanlawson:
Examples of stuff that are just hard to fix

screenshot1

screenshot2

nolanlawson:
KaiOS market share in India [seems to have declined over the past couple years](https://gs.statcounter.com/os-market-share/mobile/india).

Also, apparently they will update their Gecko version in [December of 2020](https://en.wikipedia.org/wiki/KaiOS#Release_history).

So yeah, I think I'll just wait to see what KaiOS does with PWA support this December. Right now, it's just too much work to get Pinafore running properly on old Gecko. Plus packaging up for the KaiStore is a lot of effort. Fun experiment, though.

PR #1873, opened 2020-09-18
fix: fix status bar color in iOS 14

nolanlawson:
In iOS 14, the "white" status bar option is removed: https://firt.dev/ios-14#white-status-bar-gone

But it seems that setting it to "default" does basically the same thing: https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html


nolanlawson:
On second thought, after testing on iOS 12 and 14, it looks like this change does nothing. The status bar looks the same before and after.

PR #1877, opened 2020-09-28
fix: switch from no-referrer to strict-origin

nolanlawson:
Not sending the referer is great for privacy but sadly it breaks some cross-origin scenarios, in particular when you try to run OCR on an image after a refresh. By not sending the `referer` header in the cross-origin request, many cloud providers will just not send the CORS headers (e.g. AWS Cloudfront does this).

Since the Mastodon frontend also uses `origin` (and `strict-origin` is better), I don't think we're veering too far from user expectations. Although it would be nice to be able to control outgoing link clicks and cross-origin requests separately.

### DIFF
--- a/bin/build-vercel-json.js
+++ b/bin/build-vercel-json.js
@@ -77,7 +77,7 @@ const HTML_HEADERS = {
     "form-action 'none'",
     "base-uri 'self'"
   ].join(';'),
-  'referrer-policy': 'no-referrer',
+  'referrer-policy': 'strict-origin',
   'strict-transport-security': 'max-age=15552000; includeSubDomains',
   'x-content-type-options': 'nosniff',
   'x-download-options': 'noopen',


nolanlawson:
So the impact is that now when you click a link in Pinafore, the outgoing server will know that you came from `https://pinafore.social` but it won't know anything else (e.g. path, search).

nolanlawson:
On second thought, this is somewhat-niche functionality. I can't think of any other part of the UI where we need those assets to have proper CORS headers (unlike Mastodon's audio player, for instance, ours is just `<audio>` so it doesn't need CORS). Also, the OCR feature _works_, it just doesn't necessarily work in the edge case where you upload a file, press refresh, and then try to run OCR. Maybe there is a better solution (e.g. using an IndexedDB blob cache instead of in-memory, so it can persist across refreshes).

nolanlawson:
I'm leaning towards the on-disk solution since this has privacy implications.

Krinkle:
I'd support a limited (opt-out) way to indicate that traffic came from Mastodon or Pinafore. However, I think the challenge with the current patch is that it wouldn't just indicate that, it would also indicate the specific hostname.

This provides a fairly low amount of entropy when the instance in question is using the default at `pinafore.social`. But in the case of self-hosting or less popular forks/mirrors, this not only says much about the user's profile, it could very well even identify them directly.

nolanlawson:
This is a great point. In any case I think we can keep no-referrer if I can fix the CORS issue above.

Issue #1880, opened 2020-10-22
Follow button invisible in "Dark Grayscale" theme


Krinkle:
| resting 
|--
|<img width="1095" alt="Screenshot in Dark Grayscale theme" src="https://user-images.githubusercontent.com/156867/96937722-b8b3ce80-14c0-11eb-8614-c9ad040d5f9b.png">

| hover
|--
|<img width="1098" alt="Screenshot in Dark Grayscale theme (hovering icon)" src="https://user-images.githubusercontent.com/156867/96937767-cbc69e80-14c0-11eb-9615-5d5bc00025c6.png">

For comparison, the Mastodon dark theme:

| resting | hover
|--|--
<img width="489" alt="Screenshot 2020-10-22 at 23 43 01 1" src="https://user-images.githubusercontent.com/156867/96937827-f31d6b80-14c0-11eb-960c-3daacd1469c1.png"> | <img width="490" alt="Screenshot 2020-10-22 at 23 43 01" src="https://user-images.githubusercontent.com/156867/96937829-f3b60200-14c0-11eb-8d6c-cfba4bd344c0.png">

I note that in this theme, the color does not appear to change on-hover, I'm not sure if that's intentional or not.

Krinkle:
I'd like to help fix this as my first patch. I've just gotten the local dev environment working, and now orienting myself with the code base. Could use a pointer on how to go about this :)

nolanlawson:
@Krinkle If you modify the `.scss` files in `src/scss/themes`, then reload the page, you should be able to see your changes as you make modifications. :)

PR #1881, opened 2020-10-24
fix: change dark grayscale to use have a brighter action-button color

Krinkle:
The pressed state, such as used for the "Unfollow" button, was effectively identical to the background, thus making it invisible.

Brighten it up to fix this. The relative differences are kept the same as before and match the default theme,, e.g. the brightness ranges from (darkest) Unfollow > Unfollow-hover > Follow > Follow-hover (brightest)

| | Before | After
|--|--|--
| Follow | before-follow | after-follow
| Unfollow | before-unfollow | after-unfollow

Fixes https://github.com/nolanlawson/pinafore/issues/1880.


nolanlawson:
Thank you! Looks great.

PR #1893, opened 2020-11-14
fix: fix tappable area between toolbar buttons

### LINKED ISSUE #1884: Bottom of toot appears clickable but isn't
There's an unclickable "deadzone" between buttons where the cursor is turned into a pointer for hover, but no action occurs when clicking.

I found this in context of a follow-only toot (where it additonally causes the disabled button to also appear clickable), but the issue applies to regalar/public toots as well.

capture

nolanlawson:
fixes #1884

If "disable entire toot area" is on, then the cursor becomes default between the buttons, and clicking does nothing. If it is off (default), then the cursor is always pointer and clicking between the buttons clicks the whole toot.

### DIFF
--- a/src/routes/_components/IconButton.html
+++ b/src/routes/_components/IconButton.html
@@ -21,6 +21,7 @@
     display: flex;
     align-items: center;
     justify-content: center;
+    pointer-events: auto;
   }
 
   :global(.icon-button-svg) {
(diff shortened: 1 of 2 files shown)


nolanlawson:
disabled buttons show a 🚫 at least in chrome and firefox

PR #1896, opened 2020-11-23
chore: commit vercel.json, update husky

nolanlawson:
Since switching to CircleCI, I think it makes more sense to use Vercel's built-in GitHub integration rather than having a script to publish. Our script can automatically publish `master` to `dev.pinafore.social`, and I can keep manually publishing to `pinafore.social` like before. It seems Vercel doesn't support multiple domains/subdomains for multiple branches/tags/etc.


nolanlawson:
nope, slower

PR #1897, opened 2020-11-23
chore: try concurrency 2 to speed up circleci

nolanlawson:
It's 2 vCPU, lets see if this works better


nolanlawson:
nope, slower

Issue #1693, opened 2019-12-31
Support for non-English interface language (i18n)


Krinkle:
From the README / secondary goals section. Creating a ticket for it here so as to consolidate ideas and discussion prior to offering help implement this :)

Krinkle:
Some notes:

* **Processing logic**. I imagine the main complication will be how to handle pluralisation across languages. E.g. in English one might get a way with "X new message(s)" but this gets significantly more complicated in languages where the noun varies based on whether it the number is zero, one, two, three, under 10, divisible by 11, or over 100, [etc](https://github.com/wikimedia/jquery.i18n/blob/952de9502a8d74f55975a310d4413d8aa096cc15/src/jquery.i18n.language.js#L6). For this my main experience is with using [CLDRPluralRuleParser](https://github.com/santhoshtr/CLDRPluralRuleParser). Although [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) also exists now, which has reached Stage 4 for inclusion in ECMAScript 2020, with all major browsers implementing it as of a few months ago.

* **Message format.** To substitute message parameters and handle pluralisation, we'll need to decide what syntax to use in the messages, e.g. the syntax which translators will read and write, and expose to the processing logic. For this my main experience has been with "Banana", a message format originally created for Wikipedia's backend software (MediaWiki/PHP), but now also usable through standalone JavaScript implementations (such as [jquery.i18n](https://github.com/wikimedia/jquery.i18n#message-format)). E.g. a message could look like `'Hello $1, you have $2 new {{plural:$2|message|messages}}.'`.

* **Translations.** To lower the barrier for users to contribute translations (not requiring use of GitHub or JSON etc.), we could integrate with translatewiki.net. See <https://translatewiki.net/>, <https://translatewiki.net/wiki/Translating:New_project>. This basically means they git-pull once a day to sync the English source messages, after which they automatically export any new translations that have been submitted and peer-reviewed by our translators there, via a pull request.

Krinkle:
@nolanlawson Is there something I can do as a next step toward this?

I'd also like to contribute a Dutch translation which in turn might make it easier to recommend Pinafore and Mastodon more generally toward my non-technical friends/family in the Netherlands.

nolanlawson:
[...]

PR #1904, opened 2020-11-27
fix: first stab at i18n, extract English strings, add French

nolanlawson:
This is step 1 for solving #1693. This PR doesn't really do anything except for move the English strings out into `en-US.js`.

For now, the app is still monolingual, and it doesn't support any other language. All the strings are injected into the JS and HTML at compile time.

I still haven't decided whether I want the final version to be client-rendered or server-rendered, and if I want to do `fr.pinafore.social` or `pinafore.social/fr`, but this is step 1 to internationalization at least. Let's try to get this passing the tests first and foremost.


nolanlawson:
OK cool so the tests are passing. Next up:

1. Do we do server-rendering versus client rendering?
2. If server, do we do `fr.pinafore.social` or `pinafore.social/fr`?
3. I may need to do multiple implementations and compare the performance.

nolanlawson:
Measured the perf, and `formatIntl` doesn't really add much overhead. Even on 4x slowdown, just a few milliseconds here and there. It seems trivial.

5kB added to the base bundle size, not bad.

I also added some (fairly bad) French translations, but I'm just trying to get one other language supported so I can start testing.

This will undoubtedly not be the final design, but right now this is pretty good for testing and to at least unblock people who want to add translations.

`LOCALE=fr yarn build` will build a French version.

nolanlawson:
All right, let's at least start testing this.

Issue #1925, opened 2021-01-30
format-message-parse is missing from the dependencies


Flameborn:
Package.json does not contain the *format-message-parse* package as a dependency.

*yarn add* can temporarily fix this, of course, but ideally it would be nice if this was added similarly to the other dependencies ☺️

nolanlawson:
Thank you yes, this was a mistake to add it to the devDependencies rather than the dependencies.

PR #1927, opened 2021-02-03
Add required format-message-parse dependency

Radtoo:
Trivial fix for #1925


nolanlawson:
Thanks for the PR, but it should also be removed from the devDependencies and the yarn.lock should be updated. I'll take care of this, but I appreciate the PR.

PR #1930, opened 2021-02-14
Add German locale

MarcoZehe:
This adds German as another UI language for Pinafore.


nolanlawson:
Awesome, thank you very much! FWIW there is also `emoji-picker/de.js` for translating the emoji picker, but as-is it will fall back to English for the emoji picker, no worries.

nolanlawson:
Ah, you got it! Good timing. :)

MarcoZehe:
Yes, and actually, the build will fail if that de.js for the Emoji picker isn't present. I just saw my first German language Pinafore. I'll have to correct a typo, but then I think it is ready to merge.

nolanlawson:
@MarcoZehe Hey Marco, just to check, would you be okay with me adding the emoji picker translation directly to the `emoji-picker-element` project under a different license (Apache 2)?

MarcoZehe:
Yes that's OK with me.

PR #1931, opened 2021-02-14
chore: move format-message-parse to dependencies

### LINKED ISSUE #1925: format-message-parse is missing from the dependencies

nolanlawson:
fixes #1925, see also #1927

Interestingly yarn doesn't actually need to update the yarn.lock in this case


nolanlawson:
Flaky test

PR #1935, opened 2021-02-14
fix: increase sapper timeout from 5s to 5min

### LINKED ISSUE #1924: Disable timeout during a Sapper export
When building Pinafore on a Raspberry Pi 3, the export server does not start up quick enough.

By default, Sapper defines a 5000 ms timeout when exporting, but this seems to be too short and the build fails with a timeout when crawling http://localhost:22939.

Sapper has a *--no-timeout* flag to disable this, which should allow successful builds on slower machines as well. It would be nice if this was added to the build steps.

nolanlawson:
fixes #1924

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "rollup-plugin-babel": "^4.4.0",
     "rollup-plugin-terser": "^7.0.2",
     "rtl-detect": "^1.0.2",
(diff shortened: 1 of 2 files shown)


nolanlawson:
https://github.com/nolanlawson/sapper/commit/9d2b02ad3c461d2284a9ece8aa7b402a0c8d26e8

PR #1944, opened 2021-02-15
chore: update deps

nolanlawson:


nolanlawson:
Something is messed up in cheerio, we should just get rid of it

PR #1950, opened 2021-02-15
chore: update cheerio

nolanlawson:


nolanlawson:
Looks like this bug was introduced in cheerio rc4 and fixed in master but not released to npm yet. So I'll wait for rc6 I guess.

PR #1975, opened 2021-02-28
chore: clean up input.type property access

Krinkle:
Follows-up 4218c4ce640ab33897733b4858d5c60676b4a5aa.

When accessing the IDL property, values tend to be reflected in a normalised
type and form. In the case of HTMLInputElement.type, this means the
returned value is always one of the supported and canonical lowercase
values regardless of what value the corresponding attribute holds, or
even if the attribute doesn't exist (the default will still be "text").


nolanlawson:
Nice, thank you!

PR #1991, opened 2021-03-14
perf: slightly more efficient word filter format

nolanlawson:
Makes the regexes a bit more compact, uses fewer regexes.


nolanlawson:
Related: #1990

PR #1992, opened 2021-03-14
feat: implement file share target 

### LINKED ISSUE #1009: Implement File Share Target
This would allow Pinafore to accept images shared from a camera app on Android. https://paul.kinlan.me/file-web-share-target/

nolanlawson:
fixes #1009

On Android (and I assume Chrome OS), you can now share a file with Pinafore and have it uploaded as an image/video/audio attachment. I also fixed the previous share behavior so that it preserves the "title" on Android (the "url" is not used on Android, although maybe it is on Chrome OS?).


nolanlawson:
out2

out3

PR #1996, opened 2021-03-14
perf: use scheduling.isInputPending()

nolanlawson:
See https://web.dev/isinputpending/

The easiest way to see the impact of this is to type into the compose box on 6x CPU slowdown. This technique actually slightly delays keypresses a little less, although requestIdleCallback is already doing some heavy lifting here by limiting us to <~50ms per task.

### DIFF
--- a/src/routes/_utils/scheduleIdleTask.js
+++ b/src/routes/_utils/scheduleIdleTask.js
@@ -15,9 +15,19 @@ function getRIC () {
   return typeof requestIdleCallback !== 'undefined' ? requestIdleCallback : liteRIC
 }
 
+function getIsInputPending () {
+  return process.browser && navigator.scheduling && navigator.scheduling.isInputPending
+    ? () => navigator.scheduling.isInputPending()
+    : () => false // just assume input is not pending on browsers that don't support this
+}
+
+const isInputPending = getIsInputPending()
+
 function runTasks (deadline) {
   mark('scheduleIdleTask:runTasks()')
-  while (taskQueue.length && deadline.timeRemaining() > 0) {
+  // Bail out early if our deadline has passed (probably ~50ms) or if there is input pending
+  // See https://web.dev/isinputpending/
+  while (taskQueue.length && deadline.timeRemaining() > 0 && !isInputPending()) {
     const task = taskQueue.shift()
     try {
       task()


Krinkle:
I thought that `isInputPending()` is mainly for when you have your own task scheduler, or to e.g. to consider running some tasks directly onload or in some event handler before deferring to a scheduler.

But, if one is already using `requestIdleCallback` and running inside such a handler with the `deadline.timeRemaining()` method available, I thought this would already drop to zero when e.g. input is pending. Is this not the case?

nolanlawson:
@Krinkle I'm not sure that's the case. I tested this out in Chrome anyway, and I definitely saw a difference with this patch on versus off. That seems to indicate that the "remaining time" in rIC is not influenced by whether input is pending or not. (Although it could be! Why couldn't the browser just drop this down to 0 if input is pending? Web compat?)

Krinkle:
The doc pages from Google surrounding these specs suggest that Chrome does drop `timeRemaining` to zero. What I see on <https://github.com/WICG/is-input-pending> suggests that they even considered not making this API for that reason, but that they opted for it anyway to allow for uses outside a scheduler context. E.g. onload in a background tab, inline scripts, in response to input events etc to decide to do some amount of work right away.

Looking at e.g. https://github.com/w3c/requestidlecallback/issues/72 (was the first issue I found, there may be others) Google devs mention this explicitly in relation to spec work, going back for as long as the spec has existed..

I have an old demo showcasing requestIdleCallback at <https://codepen.io/Krinkle/full/XNGEvv>. For Chrome, as soon as I double click in the output text area to select a random word flying by (and stop the scheduler to examine the output), I see that there was indeed a task who's time remaining got shortened. Also, after that interaction there's a brief periof where Chrome schedules its tasks further apart, probably to reduce chances of stuff getting in the way of other anticipated input.

For Firefox, when I click e.g. "Change View" to open/close CodePen's popover, the same happens. (It seems Firefox doesn't consider text selection as input, perhaps it is able to better separate that from other input, like scrolling nowdays if you have no relevant event handlers for those).

| Firefox | Chrome | 
|--|--
| capf | capg

Krinkle:
Anyway, all that to say, I've enjoyed using the API and it seems to work for me. I haven't compared it to `isInputPending()`. No problem with using it, I'm just curious, and mildly suspect that if using isInputPending helps, there could be a Chrome bug in here somewhere.

nolanlawson:
That's really bizarre. Maybe I only saw the impact because of the 6x CPU slowdown (some kind of bug there?). In any case, I'm using both rIC and iIP, so if `timeRemaining` drops to 0 when `iIP` is true then this PR basically does nothing.

PR #2004, opened 2021-03-18
fix: add pwa=true query param when sharing files

nolanlawson:
I'm using this (for now) as a signal that we're running in a PWA, so let's keep it consistent.


nolanlawson:
I'm not actually using this for anything, though, and I can't think what I would do differently in PWA vs non-PWA mode... but I guess I should be consistent.

nolanlawson:
> Caution: Changes to name, short_name, and start_url are not currently supported on desktop Chrome, though work is underway to support them.

https://web.dev/manifest-updates/

OK, so looks like I shouldn't change the start_url anytime soon.

PR #2005, opened 2021-03-18
feat: use web badge API to show notifications/follow requests

### LINKED ISSUE #1900: Use the new Badging API
This would make sense for notification unread counts. https://www.chromestatus.com/feature/6068482055602176

nolanlawson:
Fixes #1900

### DIFF
--- a/src/routes/_store/computations/badgeComputations.js
+++ b/src/routes/_store/computations/badgeComputations.js
@@ -25,4 +25,9 @@ export function badgeComputations (store) {
     ['numberOfFollowRequests'],
     (numberOfFollowRequests) => !!numberOfFollowRequests
   )
+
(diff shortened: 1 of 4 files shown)


nolanlawson:
Tested on Windows; it works

nolanlawson:
Capture

nolanlawson:
Looks like there's also an emoji bug on Windows in Chrome; I commented on a bug with a minimal repro for them: https://bugs.chromium.org/p/chromium/issues/detail?id=771508

nolanlawson:
For some reason this API causes Chrome to completely hang in CircleCI, so I'm just adding a check for old versions of Chrome. Hopefully this will go away when/if we upgrade the CircleCI image and the Chrome version.

PR #2014, opened 2021-03-20
Add a Wellness option to show absolute timestamps instead of relative

### LINKED ISSUE #2011: Additional wellness setting: Absolute timestamps
An interesting point that https://medium.com/swlh/user-engagement-is-code-for-addiction-a2f50d36d7ac brings up is that relative timestamps create "immediacy" - you are shown upfront how long ago something was, and can literally see time ticking away, pushing you to interact quickly with fresh content - and it also can encourage folks to wait in real time for new content (fear of missing out).

It seems straightforwards to have a toggle to display timestamps on statuses using absolute timestamps instead of calculating a relative timestamp, especially since the formatted absolute timestamp is already available (shown in the title attribute).

kepstin:
Relative timestamps can cause you to feel that things are especially
interesting because they are happening "right now"; the effect is
lessened if you see absolute timestamps instead.

This fixes #2011

As it stands, this PR has a couple of issues that still need to be solved:

When absolute timestamps are enabled, I intended to remove the `title` attribute on the timestamps, as they're no longer needed. [According to the svelte docs](https://v2.svelte.dev/guide#tags), attributes are supposed to be removed from the element if their value is `undefined` or `null`, but instead the `title` attribute is getting set to the literal string `undefined`.

The wording of the option name is a bit too long, and wraps. I'm unsure about the option group heading that I've chosen in the settings panel.

### DIFF
--- a/src/routes/_pages/settings/wellness.html
+++ b/src/routes/_pages/settings/wellness.html
@@ -33,13 +33,18 @@
     </label>
   </form>
 
-  <h2>{intl.notifications}</h2>
+  <h2>{intl.immediacy}</h2>
 
   <form class="ui-settings">
     <label class="setting-group">
       <input type="checkbox" id="choice-disable-unread-notification-counts"
              bind:checked="$disableNotificationBadge" on:change="onChange(event)">
       {intl.hideUnread}
+    </label>
+    <label class="settings-group">
+      <input type="checkbox" id="choice-disable-relative-timestamps"
+             bind:checked="$disableRelativeTimestamps" on:change="onChange(event)">
+      {intl.showAbsoluteTimestamps}
     </label>
   </form>
 
@@ -90,12 +95,14 @@
           disableReblogCounts,
           disableFavCounts,
           disableNotificationBadge,
+          disableRelativeTimestamps,
           enableGrayscale
         } = this.store.get()
         document.querySelector('#choice-check-all').checked = disableFollowerCounts &&
           disableReblogCounts &&
           disableFavCounts &&
           disableNotificationBadge &&
+          disableRelativeTimestamps &&
           enableGrayscale
       },
       onCheckAllChange (e) {
@@ -105,6 +112,7 @@
           disableReblogCounts: checked,
           disableFavCounts: checked,
           disableNotificationBadge: checked,
+          disableRelativeTimestamps: checked,
           enableGrayscale: checked
         })
         this.store.save()
--- a/src/intl/en-US.js
+++ b/src/intl/en-US.js
@@ -392,6 +392,9 @@ export default {
   hideReblogCount: 'Hide boost counts',
   hideFavoriteCount: 'Hide favorite counts',
   hideUnread: 'Hide unread notifications count (i.e. the red dot)',
+  // The quality that makes something seem important or interesting because it seems to be happening now
+  immediacy: 'Immediacy',
+  showAbsoluteTimestamps: 'Show absolute timestamps instead of relative timestamps',
   ui: 'UI',
   grayscaleMode: 'Grayscale mode',
   wellnessFooter: `These settings are partly based on guidelines from the
--- a/src/routes/_components/status/StatusRelativeDate.html
+++ b/src/routes/_components/status/StatusRelativeDate.html
@@ -4,9 +4,9 @@
    rel="prefetch"
    {tabindex}
 >
-  <time datetime={createdAtDate} title={absoluteFormattedDate}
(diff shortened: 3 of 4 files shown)


kepstin:
It looks like the `undefined` attribute problem is sveltejs/svelte#1434, which only got fixed properly in svelte v3. For now I'm just setting the `title` attribute to the empty string, which causes the tooltip to be hidden in both Chrome and Firefox.

nolanlawson:
Empty string is fine, yeah. Per HTML, the empty string is the same as undefined.

src/routes/_pages/settings/wellness.html
```diff
@@ -33,14 +33,19 @@ <h2>{intl.metrics}</h2>
     </label>
   </form>
 
-  <h2>{intl.notifications}</h2>
+  <h2>{intl.immediacy}</h2>
```
nolanlawson, line 36: Nice change; I like it. I might also go with "urgency."

nolanlawson:
Nice work, this is almost exactly the same way I would have implemented it. I'm pleased that my code was actually readable enough for you to do this! Plus being stuck on Svelte v2. :sweat_smile:

nolanlawson (approved):

nolanlawson:
Marking as "ready for review" so CircleCI can run the tests.

Should probably also write a test for this. Not sure if you were planning to do that, but I can do it if you don't have time.

nolanlawson:
Takes up quite a bit of space, especially on small screens, but I don't think it's too bad. Probably not worth bikeshedding on the timestamp format.

Screenshot from 2021-03-20 14-46-15

Compare:

Screenshot from 2021-03-20 14-46-57

nolanlawson:
Hmmm just occurred to me, though – if the point is to reduce immediacy, maybe the timestamp should be removed altogether? It's still there if you click on the toot itself, but (kinda like the boost count and fav count) you have to dig to find it. It's not immediately front-and-center.

This would solve the problem of the names being obscured by a long timestamp. I admit I haven't thought deeply about it, though.

src/routes/_pages/settings/wellness.html
```diff
@@ -33,14 +33,19 @@ <h2>{intl.metrics}</h2>
...
     <label class="setting-group">
       <input type="checkbox" id="choice-disable-unread-notification-counts"
              bind:checked="$disableNotificationBadge" on:change="onChange(event)">
       {intl.hideUnread}
     </label>
+    <label class="settings-group">
+      <input type="checkbox" id="choice-disable-relative-timestamps"
+             bind:checked="$disableRelativeTimestamps" on:change="onChange(event)">
+      {intl.showAbsoluteTimestamps}
+    </label>
```
nolanlawson, line 48: `flushChangesToCheckAll` in this file would also need to be updated.

nolanlawson (changes requested):
See comments

kepstin:
> Takes up quite a bit of space, especially on small screens, but I don't think it's too bad.

Yeah, one thing i was thinking might be a solution to that would be to show only the time if it's the same day, but worrying about that is an improvement that could be done later.

> Hmmm just occurred to me, though – if the point is to reduce immediacy, maybe the timestamp should be removed altogether? It's still there if you click on the toot itself

Also not a bad idea, but I think there's some circumstances where the timestamp link is the only way to click on the toot to see the details in the first place (iirc toots with image but no text might have this issue? and you can turn off clickable toots as an option in pinafore). So a substitute would have to be put in place, even if just a clickable icon or so.

src/routes/_pages/settings/wellness.html
kepstin: Done!

nolanlawson:
> and you can turn off clickable toots as an option in pinafore

Hmmm, good point. I guess it will have to stay, then.

> Yeah, one thing i was thinking might be a solution to that would be to show only the time if it's the same day, but worrying about that is an improvement that could be done later.

Agreed.

I think as a compromise, we could use a different date formatter for this view. For instance, it could shorten `September` to `Sep`. In the screenshots above it's `March`, but clearly for `September` it would be even more huge.

Maybe the format could look something like this?

```
Mar 20, 2021
```

So:

- three-letter format for months
- no timestamp at all. If we don't care about immediacy, then we probably don't care about the time of day, right? Just the date is enough context.

I'll whip up a quick prototype.

nolanlawson:
@kepstin I am happy to merge this with the changes from https://github.com/nolanlawson/pinafore/pull/2021. Thanks very much for opening the PR and the issue!

nolanlawson (approved):

PR #2015, opened 2021-03-20
perf: use longer-lived cache for static assets

nolanlawson:
We're getting dinged on webpagetest for [cache static assets](https://webpagetest.org/result/210320_XiG0_3c30437fd25ffd39d7028a360786804f/). I don't think this is a huge deal, since we cache everything in a Service Worker, but maybe at the CDN level of things it's better to have long-lived caches where possible.

This changes it so that every static asset is prefixed by `/v1`. We can manually increment this when these assets change. Or in the future, we could compute `v{whatever}` based on a has of all the static assets.

This is a bit playing-with-fire (could cache bad things indefinitely if done improperly), so I definitely want to test it out in dev before moving to prod.


nolanlawson:
Weird, not caching the icons in advance causes the splash screen icon to permanently be the small one rather than the big one, if the app is installed while offline in Chrome on Android. Guess we need to download all the app icons eagerly. :unamused:

nolanlawson:
I'm nervous about putting manifest.json in a subdirectory. Can't find anything online saying it wouldn't work, but it seems it's not commonly done. Just seems like it might fail in some unexpected PWA implementation (Samsung, Edge, Safari, who knows).

nolanlawson:
This is adding a lot of complexity just for a vanity metric on WebPageTest. I don't think Vercel is going to cache things any differently one way or the other. I'm just going to bump the time from 10 minutes to 60 minutes so WPT will stop complaining.

PR #2016, opened 2021-03-20
fix: remove explicit webpack chunk names

nolanlawson:
I can't see any reason to keep these, except in places where I'm explicitly using it to filter stuff in the service worker (e.g. polyfills).


nolanlawson:
Let Webpack automatically choose the chunk names and sizes. it's fine.

PR #2017, opened 2021-03-21
fix: increase cache time for static resources

nolanlawson:
See https://github.com/nolanlawson/pinafore/pull/2015#issuecomment-803489452


nolanlawson:
This feels like a waste of time.

PR #2019, opened 2021-03-21
feat: add PWA shortcuts for compose/notifications

### LINKED ISSUE #2012: Add app shortcuts
It would be neat to have a shortcut to compose a new toot... not sure if I can think of anything else that would be good here. https://web.dev/app-shortcuts/

nolanlawson:
Fixes #2012


nolanlawson:
Capture3

Neat!

PR #2021, opened 2021-03-21
fix: use day-only time format for wellness setting 

nolanlawson:
Tweaks to #2014 based on https://github.com/nolanlawson/pinafore/pull/2014#issuecomment-803632490.

- Shows only the month/day/year, not the time (when this wellness setting is activated)
- Moves logic into Status.html so we can also adjust the aria-label. Presumably users of screen readers who enable this setting will also not want to see the relative timestamp.

This fits better on small screens IMO:

Screenshot from 2021-03-21 11-08-40

Screenshot from 2021-03-21 11-08-53

/cc @kepstin

### DIFF
--- a/src/intl/en-US.js
+++ b/src/intl/en-US.js
@@ -394,7 +394,7 @@ export default {
   hideUnread: 'Hide unread notifications count (i.e. the red dot)',
   // The quality that makes something seem important or interesting because it seems to be happening now
   immediacy: 'Immediacy',
-  showAbsoluteTimestamps: 'Show absolute timestamps instead of relative timestamps',
+  showAbsoluteTimestamps: 'Show absolute timestamps (e.g. "March 3rd") instead of relative timestamps (e.g. "5 minutes ago")',
   ui: 'UI',
   grayscaleMode: 'Grayscale mode',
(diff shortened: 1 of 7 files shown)


mirabilos:
Please make this an option… I actually like the time but *hate* relative timestamps.

nolanlawson:
@mirabilos This is an option; you should be able to find it in the Settings. :)

mirabilos:
Nolan Lawson dixit:

***@***.*** This is an option; you should be able to find it in the
***@***.*** :)

No, the switch to absolute timestamps removes the time-of-day,
as far as I can tell anyway. (But if the ability to specify one’s
own date format we mentioned in #2168 will be implemented, it’ll
solve this, I think.)

I am impressed with Pinafore, however. More than with mobile clients
even, so far.

Thanks,
//mirabilos
-- 
08:05⎜<XTaran:#grml> mika: Does grml have an tool to read Apple
     ⎜    System Log (asl) files? :)
08:08⎜<ft:#grml> yeah. /bin/rm. ;)       08:09⎜<mrud:#grml> hexdump -C
08:31⎜<XTaran:#grml> ft, mrud: *g*

PR #2029, opened 2021-04-10
Add support for building on Windows

### LINKED ISSUE #1919: Unable to run in dev mode (Windows)
I'm trying to run in dev mode getting this error:

C:\Users\kapil\source\repos\pinafore>yarn run dev
yarn run v1.22.10
$ run-s build-template-html build-assets serve-dev
$ node -r esm ./bin/build-template-html.js
Built template.html in 1284.81ms
$ node -r esm ./bin/build-assets.js
$ run-p --race build-template-html-watch sapper-dev
$ node -r esm ./bin/build-template-html.js --watch
$ cross-env NODE_ENV=development PORT=4002 node -r esm ./node_modules/.bin/sapper dev
C:\Users\kapil\source\repos\pinafore\node_modules\.bin\sapper:2
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
          ^^^^^^^

SyntaxError: missing ) after argument list
    at Object.compileFunction (vm.js:355:18)
    at Generator.next (<anonymous>)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ERROR: "sapper-dev" exited with 1.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ERROR: "serve-dev" exited with 1.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

dstaley:
Fixes #1919 

There are two fixes in this PR:

1. Updates the path to the Sapper CLI from `./node_modules/.bin/sapper` (which is the path to the bash script to invoke the CLI) to `./node_modules/sapper/sapper`. This is *technically* not the same as what gets invoked with the bash script, but it does seem to work the same way, so I suspect it's a safe change to make.
2. Updates the test value for Webpack modules to a function that is OS file system separator aware. I bet there's a way to build these with the `RegExp` constructor and `path.sep`, but I figure this would be a slightly more legible version. This is not a strongly held opinion, so I'm totally open to alternatives here!

With these changes, I'm able to run `yarn dev` and get Pinafore running locally, and I'm able to run `yarn build` to get the built output into `__sapper__/export`.


nolanlawson:
Awesome, thank you!! This looks great to me.

PR #2055, opened 2021-06-20
fix: replace cheerio with jsdom

nolanlawson:
I'm tired of cheerio changing their API constantly, and I have a hard time understanding their documentation. At least I already know how to use JSDOM.


nolanlawson:
meh, too much effort

PR #2065, opened 2021-07-04
fix: remove performance-now module, use perf_hooks

nolanlawson:
Probably a breaking change for self-hosters because we'll have to bump the minimum required Node version to whichever versions support `perf_hooks`.


nolanlawson:
perf_hooks was added in Node v8, so probably ESM will be the bigger gating factor. https://nodejs.org/api/perf_hooks.html#perf_hooks_performance_measurement_apis

nolanlawson:
Removing the exclamation point as we've been using util.promisify for a while now and that required Node v8. https://nodejs.org/api/util.html#util_util_promisify_original

PR #2066, opened 2021-07-04
chore: update ruby and node versions in Circle CI

nolanlawson:
Mastodon only supports Node v12, but I need the latest version of Node v12 for ES modules support.


nolanlawson:
May have to handle this in #2064

PR #2068, opened 2021-07-05
test: skip unnecessary navigations in readonly tests

nolanlawson:
There's no point in navigating to the non-logged-in homepage if we're just going to log in afterwards.


nolanlawson:
Seems to make the tests more unstable, not worth it

PR #2103, opened 2022-01-17
Show image description if present

cincodenada:
This is somewhat a proof-of-concept, it's a pretty basic implementation but it does the job!

I'm not actually sure it's worth using the lefthand button slot, it may make more sense to just have tapping on the description expand it instead - we could add something to indicate it was tappable? I'm not sure what that would be (Tusky has a little line in the top center and an outline indicating a "drawer" of sorts, for instance)

This also shows the first line of the image description by default, which takes up a little more vertical space, but I think it's worthwhile - making image descriptions more visible in general is a net good, I think. The button showing up also would indicate this, but for short descriptions the user wouldn't have to do anything.

I also considered making a three-state button, where it would start as hidden, and tapping would toggle through hidden->first line->full, but that seems not really an improvement over just hidden->full.

I made the max height of the expanded box shove the image out of the way and fill up to half the screen, and scroll after that. Another option would be to have it expand into an overlay over the image and take up as much height as it needs - if someone is reading the description, they're generally fine not looking at the image for a bit - but that was harder to implement, so I went with the easy way for the proof-of-concept.

### DIFF
--- a/src/routes/_components/dialog/components/MediaDialog.html
+++ b/src/routes/_components/dialog/components/MediaDialog.html
@@ -25,8 +25,24 @@
         </li>
       {/each}
     </ul>
+    {#if hasDescription }
+    <div class="media-description-outside {showFullImageDescription ? 'expanded' : ''}">
+      {mediaItem.description}
+    </div>
+    {/if}
     <div class="media-controls-outside" on:click="onMediaControlsClick(event)">
-      {#if canPinchZoom}
+      {#if hasDescription }
+        <IconButton
+          className="media-control-button"
+          svgClassName="media-control-button-svg"
+          pressable={hasDescription}
+          pressed={showFullImageDescription}
+          pressedLabel="{intl.hideFullImageDescription}"
+          href="#fa-info-circle"
+          label="{intl.showFullImageDescription}"
+          on:click="toggleFullImageDescription()"
+        />
+      {:elseif canPinchZoom }
         <IconButton
           className="media-control-button media-control-button-dummy-spacer"
           svgClassName="media-control-button-svg"
@@ -77,16 +93,25 @@
             </li>
         </ul>
       {/if}
-      {#if canPinchZoom}
+      {#if canPinchZoom }
         <IconButton
           className="media-control-button"
           svgClassName="media-control-button-svg"
-          pressable={true}
+          pressable={canPinchZoom}
           pressed={pinchZoomMode}
           label="{intl.enterPinchZoom}"
           pressedLabel="{intl.exitPinchZoom}"
           href="#fa-search"
           on:click="togglePinchZoomMode()"
+          ariaHidden={!canPinchZoom}
+        />
+      {:elseif hasDescription }
+        <IconButton
+          className="media-control-button media-control-button-dummy-spacer"
+          svgClassName="media-control-button-svg"
+          href="#fa-search"
+          label=""
+          ariaHidden={true}
         />
       {/if}
     </div>
@@ -140,6 +165,20 @@
     height: calc(100% - 20px); /* 15px padding top + 5px padding bottom */
     width: calc(100% - 10px); /* 5px padding left + 5px padding right */
     padding: 15px 5px 5px 5px;
+  }
+
+  .media-description-outside {
+    margin: 5px 15px;
+    max-height: 1.5em;
+    text-align: center;
+    color: white;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+  }
+  .media-description-outside.expanded {
+    max-height: none;
+    white-space: normal;
   }
 
   .media-controls-outside {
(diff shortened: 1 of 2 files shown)


cincodenada:
Oh dang, the auto-deploy to preview is _really_ slick! Many props to whoever got that set up!

One thing I remembered after checking out the preview: an awkward bit (that would be alleviated by having the text itself be the button) is that a button still shows up even if the text is only one line long, and clicking on it effectively does nothing. I don't know if there's a great way around this - can we determine at render time if the textbox is overflowing somehow? And even if we could, we'd have to re-check if the user zoomed or resized or what have you...seems maybe more complicated than it's worth.

src/routes/_components/dialog/components/MediaDialog.html
```diff
@@ -25,8 +25,24 @@
...
+      {mediaItem.description}
+    </div>
+    {/if}
     <div class="media-controls-outside" on:click="onMediaControlsClick(event)">
-      {#if canPinchZoom}
+      {#if hasDescription }
+        <IconButton
+          className="media-control-button"
+          svgClassName="media-control-button-svg"
+          pressable={hasDescription}
```
nolanlawson, line 38: This `pressable={hasDescription}` is unnecessary; we're already in an `if hasDescription` block.

```suggestion
          pressable={true}
```

src/routes/_components/dialog/components/MediaDialog.html
```diff
@@ -77,16 +93,25 @@
             </li>
         </ul>
       {/if}
-      {#if canPinchZoom}
+      {#if canPinchZoom }
         <IconButton
           className="media-control-button"
           svgClassName="media-control-button-svg"
-          pressable={true}
+          pressable={canPinchZoom}
```
nolanlawson, line 100: Same here, this can just be `pressable={true}`.

```suggestion
          pressable={true}
```

src/routes/_components/dialog/components/MediaDialog.html
```diff
@@ -77,16 +93,25 @@
...
           pressed={pinchZoomMode}
           label="{intl.enterPinchZoom}"
           pressedLabel="{intl.exitPinchZoom}"
           href="#fa-search"
           on:click="togglePinchZoomMode()"
+          ariaHidden={!canPinchZoom}
```
nolanlawson, line 106:
```suggestion
          ariaHidden={false}
```

src/routes/_components/dialog/components/MediaDialog.html
```diff
@@ -263,12 +302,14 @@
...
     data: () => ({
-      pinchZoomMode: false
+      pinchZoomMode: false,
+      showFullImageDescription: false
     }),
     computed: {
       length: ({ mediaItems }) => mediaItems.length,
       dots: ({ length }) => times(length, i => ({ i })),
       canPinchZoom: ({ mediaItems }) => !mediaItems.some(media => ['video', 'audio'].includes(media.type)),
+      hasDescription: ({ mediaItem }) => mediaItem.description,
```
nolanlawson, line 312: A bit more clear that this is a boolean:

```suggestion
      hasDescription: ({ mediaItem }) => !!mediaItem.description,
```

nolanlawson (changes requested):
Thanks for putting together this PR. I had some minor nitpicks.

Aside from that, I think the general idea is good – there's room on the left for the :information_source: icon thanks to the pinch-zoom icon, so there's a pleasant symmetry by adding it. I also think it's a good idea to hide it when the description is missing.

As for the placement of the text itself, I would somewhat prefer for larger text center-positioned on top of the entire image. (Or maybe positioned over the bottom third, kind of like a film subtitle.) The current position seems a bit off to me.

Also I'm not sure if there's something wrong with how I'm using it, but it seems like the info toggle button doesn't actually work. The description seems always visible, whether the button is toggled or not.

Screenshot from 2022-01-17 09-25-25

Thanks again for your work on this!

src/routes/_components/dialog/components/MediaDialog.html
cincodenada: Oh! Yeah good catch - this is left over from an earlier iteration that didn't have the if blocks here.

src/routes/_components/dialog/components/MediaDialog.html
cincodenada: I can just leave `ariaHidden` off completely here, yeah?

src/routes/_components/dialog/components/MediaDialog.html
cincodenada: Good call - I slightly prefer `Boolean()` in these situations, but I think either is fine if you have a style preference.

src/routes/_components/dialog/components/MediaDialog.html
nolanlawson: Yep.

src/routes/_components/dialog/components/MediaDialog.html
nolanlawson: `Boolean()` is fine too, yeah.

PR #2108, opened 2022-02-17
Fix ios white status bar + add iOS splash screen

hellojaccc:
It'll fix ios statusbar going white.


hellojaccc:
Finally! image

It's actually a lil bit darker than color but better than white status bar imo. 

Thank you for setting up vercel, it rly helped me test things.

hellojaccc:
Yay splashscreen is also working :)

src/build/template.html
```diff
@@ -2,7 +2,7 @@
 <html lang="{process.env.LOCALE}" dir="{process.env.LOCALE_DIRECTION}">
 <head>
   <meta charset='utf-8' >
-  <meta name="viewport" content="width=device-width, viewport-fit=cover">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
nolanlawson, line 5: My understanding is that `initial-scale=1` doesn't do anything anymore. But it doesn't hurt to be explicit I guess.

src/build/template.html
```diff
@@ -15,8 +15,20 @@
        https://developers.google.com/web/fundamentals/native-hardware/fullscreen/ -->
   <meta name="mobile-web-app-capable" content="yes" >
   <meta name="apple-mobile-web-app-capable" content="yes">
-  <meta name="apple-mobile-web-app-title" content="{intl.appName}" >
-  <meta name="apple-mobile-web-app-status-bar-style" content="default" >
+  <meta name="apple-mobile-web-app-title" content="{intl.appName}">
+  <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
+  <meta name="theme-color" content="#4169e1">
```
nolanlawson, line 20: This is already handled above. (See `theThemeColor`.)

src/build/template.html
```diff
@@ -15,8 +15,20 @@
        https://developers.google.com/web/fundamentals/native-hardware/fullscreen/ -->
   <meta name="mobile-web-app-capable" content="yes" >
   <meta name="apple-mobile-web-app-capable" content="yes">
-  <meta name="apple-mobile-web-app-title" content="{intl.appName}" >
-  <meta name="apple-mobile-web-app-status-bar-style" content="default" >
+  <meta name="apple-mobile-web-app-title" content="{intl.appName}">
+  <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
```
nolanlawson, line 19: Does `black-translucent` play well with dark mode vs light mode? Or only in one or the other?

src/build/template.html
```diff
@@ -15,8 +15,20 @@
...
+  <link href="/iphonex_splash.png" media="(device-width: 375px) and (device-height: 812px) and (-webkit-device-pixel-ratio: 3)" rel="apple-touch-startup-image" />
+  <link href="/iphonexr_splash.png" media="(device-width: 414px) and (device-height: 896px) and (-webkit-device-pixel-ratio: 2)" rel="apple-touch-startup-image" />
+  <link href="/iphonexsmax_splash.png" media="(device-width: 414px) and (device-height: 896px) and (-webkit-device-pixel-ratio: 3)" rel="apple-touch-startup-image" />
+  <link href="/ipad_splash.png" media="(device-width: 768px) and (device-height: 1024px) and (-webkit-device-pixel-ratio: 2)" rel="apple-touch-startup-image" />
+  <link href="/ipadpro1_splash.png" media="(device-width: 834px) and (device-height: 1112px) and (-webkit-device-pixel-ratio: 2)" rel="apple-touch-startup-image" />
+  <link href="/ipadpro3_splash.png" media="(device-width: 834px) and (device-height: 1194px) and (-webkit-device-pixel-ratio: 2)" rel="apple-touch-startup-image" />
+  <link href="/ipadpro2_splash.png" media="(device-width: 1024px) and (device-height: 1366px) and (-webkit-device-pixel-ratio: 2)" rel="apple-touch-startup-image" />
```
nolanlawson, line 31: I'm surprised iOS needs so many splashscreens. Is there documentation somewhere for this?

nolanlawson:
Thank you for the PR! I just have a few questions about this. I'm not super familiar with iOS peculiarities, so I appreciate learning more about it. :slightly_smiling_face:

src/build/template.html
hellojaccc: https://appsco.pe/developer/splash-screens Too bad that we can't just use one image :( Will adding splashscreen affect build time/performance?

src/build/template.html
hellojaccc: Yes it works in both scenarios, someone in apple decided to name it black-translucent. It's actually translucent.

src/build/template.html
hellojaccc: Thanks! I'll remove it

src/build/template.html
hellojaccc: I removed `initial-scale=1`, and it worked great! Thanks for reviewing :) `viewport-fit=cover` seems to be problematic on ios.

src/build/template.html
nolanlawson: Yes, the service worker will eagerly download all images when the site is first loaded. We should add a filter here:

https://github.com/nolanlawson/pinafore/blob/5ecf8b8ab959698c64432eda785b0f2774576635/src/service-worker.js#L29-L36

For instance, we could add a test for the string "splash".

src/build/template.html
nolanlawson: Yes, the service worker will eagerly download all images when the site is first loaded. We should add a filter here:

https://github.com/nolanlawson/pinafore/blob/5ecf8b8ab959698c64432eda785b0f2774576635/src/service-worker.js#L29-L36

For instance, we could add a test for the string "splash".

src/build/template.html
```diff
@@ -2,7 +2,7 @@
 <html lang="{process.env.LOCALE}" dir="{process.env.LOCALE_DIRECTION}">
 <head>
   <meta charset='utf-8' >
-  <meta name="viewport" content="width=device-width, viewport-fit=cover">
+  <meta name="viewport" content="width=device-width">
```
nolanlawson, line 5: `viewport-fit=cover` was added in this PR (https://github.com/nolanlawson/pinafore/pull/1199) to fix design issues with iPhones that have a notch for the camera. Why are we removing it?

hellojaccc: I was experimenting with status bar values, removing that code fixed status bar disappearing.

I only tested on home button iPhone, so it may require testing on notched iphone.

It's been 3 years though. And i think that bug's been fixed since another projects don't have it/and works well :)

src/build/template.html
hellojaccc: Thanks! can this be marked as resolved now?

src/build/template.html
nolanlawson: I tried to go back through old issues to figure out what this does (https://github.com/nolanlawson/pinafore/pull/1199) (https://github.com/nolanlawson/pinafore/pull/1197#issuecomment-490758966), and it looks like [Apple doesn't recommend it anymore](https://stackoverflow.com/a/60939741). It looks like this was designed to fix the Snackbar (the thing that slides up from the bottom when there's an update). I don't have a notched iPhone that I can test, but I guess I can grab an iOS Simulator to try to see what this is doing.

src/build/template.html
nolanlawson: `service-worker.js` should still be updated to filter out these splashscreen files.

src/build/template.html
nolanlawson: Just tested, and I don't see any difference with or without the `viewport-fit=cover`, on notched or non-notched phones. So yeah I think we can remove this.

<details><summary>screenshots of various iphones</summary>


Screenshot from 2022-04-10 09-25-08
Screenshot from 2022-04-10 09-26-34
Screenshot from 2022-04-10 09-27-54
Screenshot from 2022-04-10 09-28-42
Screenshot from 2022-04-10 09-30-28
Screenshot from 2022-04-10 09-22-17

</details>

nolanlawson:
Thanks for the PR! I fixed it up a bit and I think it's ready to go now.

PR #2117, opened 2022-03-27
fix: add country flag emoji on windows

nolanlawson:
This is a neat solution to the problem of country flag emoji on Windows: https://www.npmjs.com/package/country-flag-emoji-polyfill


nolanlawson:
Actually, this is way harder to fix than I thought. It's almost impossible to get this to actually work everywhere in the UI.

PR #2123, opened 2022-04-10
fix: make ios status bar default color again

nolanlawson:
I don't think the change made to the status bar in #2108 is quite right. The status bar now has a translucent background when I test in iOS 15 on an iPod Touch. The timeline content bleeds through, which looks really weird.

I think this should be switched back until we figure out what magic string iOS wants.
IMG_20220410_105546

### DIFF
--- a/src/build/template.html
+++ b/src/build/template.html
@@ -16,7 +16,7 @@
   <meta name="mobile-web-app-capable" content="yes" >
   <meta name="apple-mobile-web-app-capable" content="yes">
   <meta name="apple-mobile-web-app-title" content="{intl.appName}">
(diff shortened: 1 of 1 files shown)


nolanlawson:
This only repros when adding the app to the homescreen... no repo when using Safari. :man_facepalming: 


In Safari, the background of the status bar is the blue theme color, which is correct.

FYI @hellojaccc

nolanlawson:
I can repro without even logging in. Just adding the app to the homescreen is sufficient.

Before this PR:

Screenshot from 2022-04-10 11-16-41


After this PR:

Screenshot from 2022-04-10 11-16-46

Note the unreadable text in the status bar in the first screenshot.

PR #2126, opened 2022-04-19
feat: center navbar for widths >991px

### LINKED ISSUE #403: Proposal: Center navigation on bigger screens
At the moment on bigger screen's the navigation is far away from the main timeline.

We could consider centering this, what do you think?

Minimum would be to center align the items, or could do more:

image

rylancates:
resolves #403

### DIFF
--- a/src/routes/_components/Nav.html
+++ b/src/routes/_components/Nav.html
@@ -48,7 +48,7 @@
     }
   }
 
-  @media (min-width: 992px) {
+  @media (min-width: 1400px) {
     .main-nav-ul {
       justify-content: center;
     }
--- a/src/routes/_components/Nav.html
+++ b/src/routes/_components/Nav.html
@@ -48,6 +48,12 @@
     }
   }
 
+  @media (min-width: 992px) {
+    .main-nav-ul {
+      justify-content: center;
+    }
+  }
+
 </style>
 <script>
   import NavItem from './NavItem.html'


rylancates:
smoke tested using the vercel deployment - the media query breakpoint works as expected! :smile:

src/routes/_components/Nav.html
```diff
@@ -48,6 +48,12 @@
     }
   }
 
+  @media (min-width: 992px) {
```
nolanlawson, line 51: After playing with this locally, I would like to tweak this a bit:

```suggestion
  @media (min-width: 1400px) {
```

(I'm following [Bootstrap 5's breakpoints](https://getbootstrap.com/docs/5.0/layout/breakpoints/) BTW.)

The reason is that at 992, it is always left-justified once it gets large enough to show the text rather than the icons. I feel that the center justification is only really necessary on very large screens where it starts to look odd to have so much empty space on the right side of the nav bar.

nolanlawson:
Thanks for opening the PR! I had a small tweak, but this LGTM overall. Appreciate it!

nolanlawson:
@rylancates I've been playing around with this on `dev.pinafore.social` and I have to admit... I kinda just don't like it. :stuck_out_tongue: Maybe it's muscle memory or something, but my mouse cursor always seems to want to go to the upper-left than the center.

How would you feel about making this optional? We can set the breakpoint back to 991px in that case, because the user has indicated that they want it centered.

rylancates:
Sure, I'm more than happy to make this optional! To be honest it's also a win for me as my use case includes less than the breakpoint size you'd pondered adjusting to.

I'll take a look at this over the course of the next week 😃

PR #2128, opened 2022-04-26
fix: make the center nav optional

nolanlawson:
As discussed in https://github.com/nolanlawson/pinafore/pull/2126#issuecomment-1109161943 I'd like for centering the nav header to be optional. It just looks odd to me personally.

The reason for this convoluted system for toggling the style is to avoid a flash of unstyled content when reloading.


nolanlawson:
FYI @rylancates

rylancates:
Ah, just seeing that you actually implemented. Thank you!

PR #2138, opened 2022-05-01
fix: fix votes count on Misskey polls

nolanlawson:
Misskey reports `voters_cont` as null, so just use the `votes_count` in that case.

The percentages won't be as a percent of the number of voters – instead it'll be as a percent of the number of individual votes. But this is fine. Maybe Misskey only does this for non-multiple-choice polls.

The poll looks like this in Misskey:

```js
    "poll": {
        "id": "96327",
        "expires_at": "2022-05-08T08:26:05.336Z",
        "expired": false,
        "multiple": false,
        "votes_count": 123,
        "voters_count": null,
        "voted": false,
        "own_votes": [],
        "options": [
          /* ... */
        ],
        "emojis": []
    }
}
```

### DIFF
--- a/src/routes/_components/status/StatusPoll.html
+++ b/src/routes/_components/status/StatusPoll.html
@@ -292,13 +292,16 @@
     computed: {
       pollId: ({ originalStatus }) => originalStatus.poll.id,
       poll: ({ originalStatus, $polls, pollId }) => $polls[pollId] || originalStatus.poll,
-      options: ({ poll, originalStatusEmojis, $autoplayGifs, votersCount }) => (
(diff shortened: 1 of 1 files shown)


nolanlawson:
Interesting, nope, even for multiple-choice the `voters_count` is null. Ah well.

nolanlawson:
I can't find an open issue on Misskey's GitHub. It looks like their API doesn't report either voters count or votes count. I'm guessing that when it federates to Mastodon, it just counts up the votes as votes count and ignores voters count. Well, whatever. 

https://github.com/mastodon/mastodon/pull/11917

PR #2140, opened 2022-05-06
fix: use empty array for ws protocols instead of null

tsmethurst:
Heya! I was running into a bug using Pinafore with GoToSocial, but only under Chrome.

Specifically, was getting the following error:

`Error during WebSocket handshake: Sent non-empty 'Sec-WebSocket-Protocol' header but no response was received`

I looked at the request headers that were being sent by the websocket connection, and saw 'Sec-Websocket-Protocol: null`.

After a bit of digging through the code, I noticed that if you pass 'null' to the WebSocket client constructor, it will pass 'null' via the headers, whereas if you just pass an empty array by default, this header is skipped.

This seems to resolve the issue :) Tested with Chrome and GtS


nolanlawson:
Thanks for reaching out. Yeah looks like I should either be omitting the parameter or using the empty array: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket

> If it is omitted, an empty array is used by default, i.e. `[]`.

I'm surprised this hasn't caused problems before, but looks like a good fix. Thanks!

tsmethurst:
No problem! Thanks for pinafore 😍

Sent from ProtonMail mobile

-------- Original Message --------

PR #2204, opened 2022-11-19
chore: update yarn.lock, check lockfile in CI

nolanlawson:


.circleci/config.yml
```diff
@@ -165,7 +165,7 @@ jobs:
       - restore_yarn_cache
       - run:
           name: Yarn install
-          command: yarn install --immutable
+          command: yarn install --frozen-lockfile
```
nolanlawson, line 168: yarn v1 vs v2+ :facepalm:

PR #2164, opened 2022-11-01
feat(ui): bottom nav

bennypowers:
Closes #202 

I had some difficulty testing locally

```
$ yarn run-mastodon
yarn run v1.22.19
$ node ./bin/run-mastodon.js
Installing Mastodon...
Setting up mastodon database...
ChildProcessError: Command failed: createdb -h 127.0.0.1 -U pinafore -w pinafore_development
createdb: error: connection to server at "127.0.0.1", port 5432 failed: FATAL:  Ident authentication failed for user "pinafore"
 `createdb -h 127.0.0.1 -U pinafore -w pinafore_development` (exited with error code 1)
    at callback (/home/bennyp/Developer/pinafore/node_modules/child-process-promise/lib/index.js:33:27)
    at ChildProcess.exithandler (node:child_process:406:5)
    at ChildProcess.emit (node:events:527:28)
    at maybeClose (node:internal/child_process:1092:16)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:302:5) {
  code: 1,
  childProcess: {
    _forkChild: [Function: _forkChild],
    ChildProcess: [Function: ChildProcess],
    exec: [Function: exec],
    execFile: [Function: execFile],
    execFileSync: [Function: execFileSync],
    execSync: [Function: execSync],
    fork: [Function: fork],
    spawn: [Function: spawn],
    spawnSync: [Function: spawnSync]
  },
  stdout: '',
  stderr: 'createdb: error: connection to server at "127.0.0.1", port 5432 failed: FATAL:  Ident authentication failed for user "pinafore"\n'
}
error Command failed with exit code 1.
```

that was after installing enabling and starting postgres, then `CREATE USER mastodon CREATEDB;` and `CREATE USER pinafore CREATEDB;`

:shrug:

PR #2205, opened 2022-11-19
feat(ui): bottom nav

nolanlawson:
Same as #2164 but with some tweaks on my side, and plus we're running the CI tests now because it's my branch.


src/routes/_store/store.js
```diff
@@ -14,6 +14,7 @@ const persistedState = {
   currentRegisteredInstance: undefined,
   // we disable scrollbars by default on iOS
   disableCustomScrollbars: process.browser && /iP(?:hone|ad|od)/.test(navigator.userAgent),
+  bottomNav: false,
```
nolanlawson, line 17: Considering turning this on for all phone users (iOS and Android), but seems really tricky to detect properly. Folks can just turn this on themselves if they like.

Issue #2159, opened 2022-10-02
Accessibility: low contrast icons in pitch black theme


davidbtc2009:
The icons and character count below the compose text box are extremely low contrast when in pitch black theme.

Edit: actually, this applies to the reply, boost, fav, and more icons below posts as well.

nolanlawson:
Some of the themes are contributed, and not all of them are super accessible. Thanks for reporting.

asgrim:
I'd say the contrast isn't great even on the default light theme ("Royal").

![Screenshot from 2022-11-07 16-25-47](https://user-images.githubusercontent.com/496145/200362479-fe4f30e8-3ec9-4f03-af62-4748da878656.png)

When scrolling through a feed, I have to look quite hard to tell if I've starred a post.

nolanlawson:
@asgrim I'm starting to think than an explicit support for high-contrast mode is the best approach.

The CSS `prefers-contrast` did not exist when I first wrote Pinafore. Now that it exists, I should be able to make everyone happier maybe: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast

nolanlawson:
I've also seen folks complain about:

- Gecko theme
- Ozark theme

Certain themes definitely don't have this problem: Cobalt uses a separate color entirely for boost/favorite, and Oaken has a lot of contrast.

I need to find some accessibility guidelines for what the best values here would be. The foreground color also has to be light dark enough to be distinguishable against the background.

Maybe we need some other affordance entirely as an option in the settings – e.g. a prominent border, or a background color, or something like that.

nolanlawson:
Related: #1646 #1831

NickColley:
Consolidating into https://github.com/nolanlawson/pinafore/issues/2230, thank you so much for raising.

PR #2208, opened 2022-11-20
Improve colour contrast

NickColley:
Made some improvements to contrast for all themes, there's still a lot of issues but I'm thinking I can make chunks towards where we need to be.

I will update this pull request with screenshots but feel free to check it out locally if you are curious in the meantime.

Improves https://github.com/nolanlawson/pinafore/issues/2159 but does not resolve it.

### DIFF
--- a/src/routes/_components/NavItem.html
+++ b/src/routes/_components/NavItem.html
@@ -27,7 +27,7 @@
   }
 
   .nav-icon-and-label {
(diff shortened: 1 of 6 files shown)


NickColley:
Might split this up until smaller pull requests actually but it's late, so will do that tomorrow.

PR #2209, opened 2022-11-20
Show usernames in page titles

g2p:
This shows usernames in page titles and reorders it a bit.

I find it convenient to distinguish tabs at a glance.

I have no idea what I'm doing, it may be redundant to compute this in several profile-related components, and I don't know how this works while the profile is being loaded asynchronously.

### DIFF
--- a/src/intl/de.js
+++ b/src/intl/de.js
@@ -170,12 +170,12 @@ export default {
       true {({count})}
       other {}
     }
+    {name}
+    ·
     {showInstanceName, select,
       true {{instanceName}}
       other {Pinafore}
     }
-    ·
-    {name}
   `,
   pinLabel: `{label} {pinnable, select,
     true {
--- a/src/intl/en-US.js
+++ b/src/intl/en-US.js
@@ -174,12 +174,12 @@ export default {
       true {({count})}
       other {}
     }
+    {name}
+    ·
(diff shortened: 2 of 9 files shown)


nolanlawson:
This looks reasonable to me. Thanks!

nolanlawson:
I tweaked slightly how the name appears, because it looked a bit odd to me to say `@foo Profile`. Now it says `@foo · Profile`.

And yes, it looks fine while the account is loading. You can repro by pressing refresh – it takes a bit for the profile to load. :slightly_smiling_face: 

Thanks again!

g2p:
Thank you for having a look! I'm also trying to set a title for statuses (author, possibly some ellipsized text), but I haven't figured it out yet.  I figured out statuses are shown as a mini-timeline, but I don't know how to reach inside the components.

nolanlawson:
Hm, I'm not sure what title you could put on a status? The text of the status itself?

g2p:
Author's account at least, possibly a little bit of text from the status, yes.

g2p:
I've rebased.  You can merge at this point, I don't think if I'll be able to update titles for Toot pages soon.  Also I've updated the tests since you made #2211.

nolanlawson:
Closed in #2211. Thank you!

PR #2210, opened 2022-11-20
fix: action buttons background for better contrast

NickColley:
The issue with the current buttons is we are making the distinction between pressed based on the contrast between two colours, this will always be difficult.

This proposal instead changes the shape by relying on an 'inverse' style which makes it very clear if it is pressed or not no matter someone's perception of colour.

### Status toolbar

| Before | After |
| - | - |
|  |  |

### Compose toolbar

| Before | After |
| - | - |
|  |  |

### Dark
| Before | After |
| - | - |
|  |  |

### DIFF
--- a/src/routes/_components/IconButton.html
+++ b/src/routes/_components/IconButton.html
@@ -25,10 +25,14 @@
   }
 
   :global(.icon-button-svg) {
+    --svg-bg-color: transparent;
     width: 24px;
     height: 24px;
     fill: var(--action-button-fill-color);
     pointer-events: none; /* hack for Edge */
+    border-radius: 0.1%;
+    background: var(--svg-bg-color);
+    box-shadow: 0 0 0 5px var(--svg-bg-color);
   }
 
   :global(.icon-button.big-icon .icon-button-svg) {
@@ -39,26 +43,23 @@
   /*
    * regular styles
    */
-
   :global(.icon-button:hover .icon-button-svg) {
-    fill: var(--action-button-fill-color-hover);
+    --svg-bg-color: var(--action-button-bg-color-hover);
   }
-
-  :global(.icon-button.not-pressable:active .icon-button-svg,
-  .icon-button.same-pressed:active .icon-button-svg) {
-    fill: var(--action-button-fill-color-active);
+  :global(.icon-button:active .icon-button-svg) {
+    --svg-bg-color: var(--action-button-bg-color-active);
   }
-
+  :global(.icon-button.pressed .icon-button-svg) {
+    --svg-bg-color: var(--action-button-bg-color-pressed);
+  }
+  :global(.icon-button.pressed:hover .icon-button-svg) {
+    --svg-bg-color: var(--action-button-bg-color-pressed-hover);
+  }
+  :global(.icon-button.pressed:active .icon-button-svg) {
+    --svg-bg-color: var(--action-button-bg-color-pressed-active);
+  }
   :global(.icon-button.pressed.not-same-pressed .icon-button-svg) {
     fill: var(--action-button-fill-color-pressed);
-  }
-
-  :global(.icon-button.pressed.not-same-pressed:hover .icon-button-svg) {
-    fill: var(--action-button-fill-color-pressed-hover);
-  }
-
-  :global(.icon-button.pressed.not-same-pressed:active .icon-button-svg) {
-    fill: var(--action-button-fill-color-pressed-active);
   }
 
   /*
--- a/src/scss/themes/_base.scss
+++ b/src/scss/themes/_base.scss
@@ -44,12 +44,13 @@
   --nav-svg-fill-hover: #{$secondary-text-color};
   --nav-text-color-hover: #{$secondary-text-color};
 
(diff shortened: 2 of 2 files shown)


NickColley:
cc @joelanman if you're interested

NickColley:
There are alternative approaches I tried such as having an outlined icon with no fill but would require me editing the icons themselves.

nolanlawson:
I really appreciate this PR! The contrast issue definitely needed to be fixed for a long time.

I'm not sure I'm totally on-board with this design decision, though. Admittedly I'm not a designer, and maybe it's just that I've used Pinafore for so long, but the filled background looks "off" to me. Also, some of the animations now don't look quite as seamless (e.g. the favorite star or the boost buttons).

Here is an alternative I can think of:

1. Make this an option in the settings. (See #2205 for how I handle this with CSS custom properties).
2. Enable this option by default based on [`prefers-contrast: more`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast).

This would not fix all the contrast issues (even the home page on desktop does not meet AA due to the "Home" button), and the contrast could still be improved on _many_ of the themes (I did not give them a thorough audit when I wrote them or merged PRs from people who submitted them), but it could at least "stop the bleeding" for people who can't distinguish the different colors on their chosen theme.

What do you think?

NickColley:
Cheers Nolan.

I did start doing more fixes to the contrast but there are so many I want to try and make smaller improvements over time if possible.

While your feedback on how it feels is subjective so indeed not the best design feedback perhaps it's that the filled background gives the pressed state too much of visual precedence or it's now higher in the visual hierarchy.

I think that's not really a problem for the compose toolbar. So perhaps they should be separated so we can keep a lower visual hierarchy for the toot buttons, but I think that'd have to be achieved with updates to the icons themselves really.

NickColley:
I did have a nice setup where the favourite star was outlined with no fill and then filled in when pressed but the issue is we don't have that for retweets, so we'd want to have a custom icon set to go along that road which obviously is more commitment. But I think would allow the buttons to command less attention at least...

NickColley:
Closing for now.

NickColley:
I don't think it's a good idea to have core accessibility as UI settings, the decision to put centering the navigation bar is something I'm not a fan off for example. We should instead be bold to make changes that increase usability and accessibility and accept that some people will be upset at change for changes sake (it's always the case)

nolanlawson:
Sorry for the bad feedback, you're right. Like I said, I'm not a designer, so I don't really have the language to describe my opinion. :disappointed: 

I think you're right about it commanding too much attention. I think you're also right about the :warning: icon in the compose bar being less of a big deal because it doesn't "command attention" like the buttons in the timeline, but I do also like the idea of having consistency between the two, so I'm not sure about having a separate system for each.

I guess in my mind I had imagined fixing each individual theme (or even dropping some themes – I'm not sure how popular "Cobalt" is for instance), but you're right that it's a onesy-toosy of fixing each theme individually, which would take a lot of time. (Hence why I have been lazy and haven't done it.)

> I don't think it's a good idea to have core accessibility as UI settings

I agree up to a certain point, but I think we're already going down this path with things like `prefers-reduced-motion`, `prefers-contrast: more`, or the "underline links" and "always show the focus ring" options I added in the settings. In other cases like "autoplay animated GIFs" I went the other direction and made the pro-accessibility option the default.

Maybe I'm wrong about the contrast thing (if so: educate me! I only really know about foreground-background contrast on text), but especially since people can choose their own themes, I was not under the impression that it was a huge showstopper for users. Hence why I was happy to make this PR an option in the settings. :slightly_smiling_face:

NickColley:
Basically you want to have "enough contrast" as your default. And "high contrast" as an optional extra.

Some users find high contrast causes visual stress, so it's highly dependent on people's needs.

When making design decisions we should aim to meet the needs of as many people by default, and then only after that include options to increase this further.

It is not ethical to instead have a segregated by default design approach where we suggest accessible design should always be behind a flag.

It's okay you don't have the words for it I think I understood you anyway. 👍

NickColley:
But the contrast most of the themes are at at the moment will be show stopper for anyone who has a visual impairment. But even myself with corrected vision I find it difficult to know when I've boosted/favourited.

nolanlawson:
Understood, and I agree with you in principle. Some of the themes are tricky for me as well! I use "Majesty" by default which seems pretty good (to me, subjectively).

Are there any accessibility guidelines for contrasting the color of two button states? (Other than "Don't use color to communicate differences," which I see a lot in the a11y literature. :slightly_smiling_face:) I did a quick search of the web and did find anything. I only see [this blurb in WCAG](https://www.w3.org/TR/WCAG21/#contrast-minimum) about adjacent colors, which I think doesn't apply here.

NickColley:
We'd want the colour contrast difference between the two states to be above 3:1 imo which is sometimes easier said than done.

This is why it makes more sense for us to solve this problem by changing the shape of the icons themselves.

I think the solution that'll be the most elegant is to increase the stroke of the icon or have an outlined icon that is then filled in.

Trying to use colour to communicate state is a losing game really.

PR #2218, opened 2022-11-21
Create ru-RU.JS

vorons:
Russian language


nolanlawson:
Also FWIW, the emoji picker currently does not have a Russian translation: https://github.com/nolanlawson/emoji-picker-element/tree/master/src/picker/i18n

PR #2227, opened 2022-11-23
fix: Focus the textarea in the Compose dialog instead of the Close button.

### LINKED ISSUE #1839: Move keyboard focus to the text area when composing via the dedicated Compose dialog
## Steps to reproduce:

* When the main timeline is in focus, hit *c*, or *m*.

## Result

Keyboard focus moves to the dialog's *close* button.

## Expected result

Keyboard focus should move to the composer text area.

jcsteh:
ComposeBox already specified autoFocus for the ComposeInput. However, this didn't work because the dialog was disabled when the ComposeInput code tried to focus the textarea. To fix this, tweak A11yDialog to look for the autofocus attribute when determining what to focus. This is consistent with the behavior for native HTML dialogs. Then, have ComposeInput set this attribute if it's in a dialog. Fixes #1839.

### DIFF
--- a/src/routes/_components/compose/ComposeInput.html
+++ b/src/routes/_components/compose/ComposeInput.html
@@ -117,7 +117,15 @@
             firstTime = false
             const { autoFocus } = this.get()
             if (autoFocus) {
-              requestAnimationFrame(() => textarea.focus({ preventScroll: true }))
+              const { realm } = this.get()
+              if (realm === 'dialog') {
+                // If we're in a dialog, the dialog will be hidden at this
+                // point. Also, the dialog has its own initial focus behavior.
+                // Tell the dialog to focus the textarea.
+                textarea.setAttribute('autofocus', true)
+              } else {
+                requestAnimationFrame(() => textarea.focus({ preventScroll: true }))
+              }
             }
           }
         })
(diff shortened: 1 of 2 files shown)


jcsteh:
The other thing to flag is that it'd probably be nicer if a particular dialog could just specify the element it wants to focus, but I wasn't able to work out a way to identify a particular component and a specific element inside that component.

MarcoZehe:
This works great, just tested it on the deployment.

nolanlawson:
Awesome, thank you! Making the change directly on a11y-dialog makes the most sense to me.

FWIW, I had done a similar thing for the emoji picker, but I'm doing it manually in `EmojiDialog.html`:

https://github.com/nolanlawson/pinafore/blob/6b1533c9476de9f098c2c66c2a7a0f5acdf36b00/src/routes/_components/dialog/components/EmojiDialog.html#L90-L96

Arguably I should use the same system you're using here – although incidentally, this code will not work because the emoji picker is inside of shadow DOM. So maybe it's best to just keep that as-is for now.

nolanlawson (approved):

jcsteh:
> Arguably I should use the same system you're using here – although incidentally, this code will not work because the emoji picker is inside of shadow DOM.

As I understand it, A11yDialog's getFocusableChildren function (which is used by setInitialFocus) looks into shadow roots, so I think this autofocus tweak would work here. Technically, that wouldn't work in a native dialog - we'd have to delegate focus to the host - but it would allow you to use a common code path here at least if you wanted to.

nolanlawson:
It looks into shadow roots specifically for the emoji picker, yeah. I added that as a hack to make shadow DOM work for that one component. There are no other shadow roots in Pinafore.

https://github.com/nolanlawson/pinafore/blob/a775bd9193a8d3b0f24201d09667943210a3e1f4/src/routes/_thirdparty/a11y-dialog/a11y-dialog.js#L372-L377

PR #2229, opened 2022-11-23
fix: If the focus is inside a toot, make keyboard commands operate relative to the parent toot.

### LINKED ISSUE #2228: If a control inside a toot is focused, arrow keys should still move relative to that toot
Steps to reproduce:

1. Focus the first toot in a timeline.
2. Press tab to focus a control inside it.
3. Press down arrow.
    - Expected: The second toot should be focused.
    - Actual: The first toot is focused.

This is inconvenient if:

1. A screen reader user uses their review cursor to more closely inspect the content of a toot and inadvertently focuses a control within.
2. You interact with a toot in some way that requires you to move focus inside it.

jcsteh:
Previously, if focus was on an element inside a toot instead of the toot itself (e.g. moving to a toot and pressing tab), keyboard commands acted as if no toot was active. In particular, this meant that the arrow keys scrolled to the first visible toot.

Fixes #2228.

### DIFF
--- a/src/routes/_components/shortcut/ScrollListShortcuts.html
+++ b/src/routes/_components/shortcut/ScrollListShortcuts.html
@@ -119,7 +119,13 @@
         if (!activeElement) {
           return null
         }
-        const activeItem = activeElement.getAttribute('id')
+        // The user might be focused on an element inside a toot. We want to
+        // move relative to that toot.
+        const activeArticle = activeElement.closest('article')
+        if (!activeArticle) {
+          return null
+        }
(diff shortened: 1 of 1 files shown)


MarcoZehe:
This is a really good change! Thank you, Jamie! I just tried this on the deployment for this PR, and it works beautifully.

nolanlawson:
Tested it out, works great. Opened a new PR so we can run the tests: https://github.com/nolanlawson/pinafore/pull/2232

nolanlawson (approved):

PR #2234, opened 2022-11-23
fix: dont rely on colour for boost/favourite state

NickColley:
By changing the shape it means no matter what the colour difference between pressed and non-pressed it'll be possible to know the state.

I tried having it appear after the animation and it felt nicer before but let me know what you think when you try it out.

### DIFF
--- a/src/routes/_components/IconButton.html
+++ b/src/routes/_components/IconButton.html
@@ -13,7 +13,7 @@
 >
   <SvgIcon className="icon-button-svg {svgClassName || ''}" ref:svg {href} />
   {#if checked}
-  <SvgIcon className="icon-button-svg icon-button-check" href="#fa-check" />
+    <SvgIcon className="icon-button-svg icon-button-check" ref:check href="#fa-check" />
   {/if}
 </button>
 <style>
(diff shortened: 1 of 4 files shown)


nolanlawson:
This looks beautiful to me! I just have some small change I'd like to make, which is that I think it would be neat to transition the checkmark from the old color to the new color. I've gone ahead and pushed this to your branch if you don't mind: 3bc3ba02

Eventually I would also love for the "follow" button on `AccountProfileFollow.html` to have the same checkmark, but I tried it out and the checkmark is the wrong size. So we can maybe fix that in a follow-up PR, no biggie.

There are also some test issues but I'm guessing those are flaky tests or problems with the tests, I'll follow up. Thanks!

NickColley:
Nice, I like it. Look at us go, tag team ha

nolanlawson (approved):

NickColley:
Thanks for your patience working through this, here's to more improvements... :)

asadotzler:
Thank you, Nick and Nolan. This was unexpectedly quick. (Open and participatory software projects FTW!!!) 

This will make the experience significantly better for everyone, and should be of particular value to users with certain types of low vision. I read Nolan's post on the maintenance burden, and while I'm not a programmer, so I cannot help directly, I am a product manager for Firefox accessibility (24 yrs experience with Mozilla) and I offer what ever help I can including bug reports, pre and post change testing, design advice or reviews, etc. to improve the usability of this front end for people with disabilities (and everyone else who benefits from that work.) So far I much prefer it to the other front ends I've tried and hope to see it live on. 

(Also apologies for the non-technical comment spam in your bug tracker, but I was so happy to see this update today, and as I've been expressing Thanksgiving gratitude all day in meatspace, I couldn't help myself here.)

NickColley:
@asadotzler much love to you my friend

PR #2236, opened 2022-11-23
Query remote Mastodon server for follower list, on clicking follower …

ringtailsoftware:
When looking at followers of a profile, I only see followers on my local instance - which isn't what I want.

This change:
When viewing the follower list for an account, query the API on the account's home instance not mine, to get a full list of followers (up to page limit).
On clicking the follower, go to their home instance's profile page where I can view posts and decide to follow.

### DIFF
--- a/src/routes/_api/followsAndFollowers.js
+++ b/src/routes/_api/followsAndFollowers.js
@@ -1,5 +1,7 @@
 import { get, paramsString, DEFAULT_TIMEOUT } from '../_utils/ajax.js'
 import { auth, basename } from './utils.js'
+import { getAccount } from '../_api/user.js'
+
 
 export async function getFollows (instanceName, accessToken, accountId, limit = 80) {
   let url = `${basename(instanceName)}/api/v1/accounts/${accountId}/following`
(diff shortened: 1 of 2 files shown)


nolanlawson:
Thanks for the PR! I understand where you're coming from – it is totally confusing that you may see different followers on your home instance than on the instance hosting the remote account.

However, there are a few things that make me uneasy about this:

1. This doesn't just apply to remote followers – the number of favorites, boosts, etc. and even the replies may vary from server to server.
2. In a sense, this is a Mastodon issue, not a Pinafore issue – our home Mastodon server chooses not to federate certain information (for whatever reason Mastodon has for this).
3. This PR changes the privacy/security model of Pinafore – when I log in to my instance, I would not expect that Pinafore would start making HTTP requests to arbitrary instances on the internet. I may trust my instance, but I don't trust every instance in the fediverse. It feels like this opens up potentially new attack vectors – e.g. a malicious instance could try to craft a response in such a way that it changes Pinafore's behavior. (This particular attack doesn't _seem_ possible to me, but I haven't fully thought it through.) In any case, it is also a privacy issue, because now the third-party server knows that I've looked at an account on their instance.

So overall, for these reasons, I am inclined not to approve this PR. Do you know if other Mastodon clients have taken a similar approach?

ringtailsoftware:
Thanks for the feedback, this all makes sense - I can see that my change would take pinafore in a different direction to the one you want. It's an excellent client though, so thanks.

I do not know if any other clients reach out to remote servers.

I suspect that what I'm really looking for is a tool to explore the social graph and find interesting new follows. Given the structure of Mastodon, that's more like a web crawler than a client tied to a single server.
Perhaps I'll keep working in that direction on my own fork.

Issue #2226, opened 2022-11-23
"Loading more" alert creates extraneous verbosity for screen reader users


jcsteh:
Thanks for such a fantastic, delightful Mastodon client.

Strictly speaking, the "Loading more" alert is correct a11y because it makes it clear that more content is loading, which would otherwise not be obvious. On the other hand, pragmatically speaking, I find this very distracting and intrusive as a screen reader user, as it interrupts my browsing of the timeline. I can silence it of course, but it is still jarring.

I think it would be worth considering removing this alert, though this would obviously need feedback from other screen reader users.

CC @MarcoZehe for his thoughts as another screen reader user.

nolanlawson:
@jcsteh Hmmm interesting, this wasn't specifically something I added only for screen reader users – the footer transforms into a spinner alongside the explicit visible test "Loading more." However, I imagine that 99% of the time, this text flashes for a brief moment before the network request returns. So from a screen reader's perspective it probably announces the text every time. (And for sighted users, there is an opacity animation from 0 to 1, so it's less jarring.)

I think I could fix this by adding a delay to when the `aria-hidden="false"` is set. For instance, it could wait 1 second before doing so. Or I could explicitly make it a live region with `aria-live="polite"`. My hunch is that the first one is better, but which one would you prefer?

MarcoZehe:
Does this one have a role="alert"? If so, that has an implicit aria-live="assertive" attached to it. This is what makes it jarring. If you removed the role attribute and set aria-live="polite" instead, it would still announce when loading new stuff or some similar, but be non-intrusive. Much like the "Favorited" and "Boosted" announcements you made talk recently.

jcsteh:
It does have role="alert". My gut feeling is that this should just be removed. The downside is that screen reader users wouldn't get notified that content is loading. However, given how quick this usually is, I'm not sure how useful this is. By the time the announcement has occurred, the content has probably finished loading.

aria-live="polite" might be better than role="alert", but this might still mean that the announcement occurs before the text of the newly focused item.

If the concern is that screen reader users wouldn't know about slow loads, another option is to not have role="alert" by default, but set role="alert" if the message is displayed for longer than, say, 1 second. This way, a screen reader user wouldn't get the message in most cases, but they'd know about it within a second if the load was taking too long. This obviously isn't equivalent to the sighted UX, but I think it's more pragmatic.

MarcoZehe:
I think this would be a good compromise.

[...]

PR #2240, opened 2022-11-24
fix: announce "loading more" with aria-live=polite after delay

nolanlawson:
This implements @jcsteh's suggestion here: https://github.com/nolanlawson/pinafore/issues/2226#issuecomment-1325805513

It announces "Loading more" after a 1 second delay, if the new statuses at the bottom of the timeline are delayed.

### DIFF
--- a/src/routes/_components/timeline/LoadingFooter.html
+++ b/src/routes/_components/timeline/LoadingFooter.html
@@ -65,11 +65,30 @@
   }
 </style>
 <script>
+  import { observe } from 'svelte-extras'
   import LoadingSpinner from '../LoadingSpinner.html'
   import { store } from '../../_store/store.js'
   import { fetchMoreItemsAtBottomOfTimeline } from '../../_actions/timeline.js'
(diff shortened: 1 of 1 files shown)


nolanlawson:
One thing I should probably confirm is whether the `announceAriaLivePolite` is implemented properly. I'm just seeing the `textContent` on an `aria-live="polite"`, and I'm not clearing the text if it's unchanged. This "feels" to me like it shouldn't work, but it seems to work in Firefox+NVDA and Safari+VoiceOver. So maybe it's fine?

MarcoZehe:
I believe aria-atomic is your friend here. There you can specifically tell browsers and screen readers how added, removed, or replaced text should be handled.

nolanlawson:
Thanks! I looked into both aria-atomic and aria-relevant, but based on the MDN docs, they don't apply here because the live region is a single element with one child text node. I think I will wait to see if I get a complaint from some screenreader+UA combo, and in that case I can look into fixes.

PR #2241, opened 2022-11-24
fix: use dvh for bottom nav

### LINKED ISSUE #2221: Nav bar invisible when placed at bottom on Chrome on Android
Looks like I messed something up with #2164 : https://mastodon.social/@developit/109377628451233871 

My bad for only testing in Chrome/Firefox on desktop. :stuck_out_tongue:

nolanlawson:
Fixes #2221

Uses 100dvh instead of 100vh to position the bottom nav. I'm using a polyfill because Chrome [does not support it yet](https://caniuse.com/viewport-unit-variants) (but it will next release). Plus, there's old Safari to worry about. The polyfill [looks reasonable](https://github.com/joppuyo/large-small-dynamic-viewport-units-polyfill/blob/master/src/index.js).

This looks pretty janky in every browser I tested (Firefox & Chrome on Android, old Safari/iOS and new Safari/iOS), but I guess it's better than the nav being hidden. It is weird though that it doesn't smoothly reposition as you scroll – instead, the nav kind of jumps around as the URL bar is shown or hidden. Maybe Chrome 108 will fix that.

Who knew it was so hard to position something on the bottom of the screen?

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,6 @@
     "format-message-parse": "^6.2.4",
     "glob": "^7.2.0",
     "is-emoji-supported": "^0.0.5",
-    "large-small-dynamic-viewport-units-polyfill": "^0.0.4",
     "li": "^1.3.0",
     "localstorage-memory": "^1.0.3",
     "mkdirp": "^1.0.4",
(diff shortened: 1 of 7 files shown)


nolanlawson:
Based on Edge Beta on Android, Chrom(ium) 108 does not seem to fix that. :disappointed:

PR #2246, opened 2022-11-25
fix: bring check closer to icon

NickColley:
Just a small aesthetic change to make them feel more connected.

## Screenshot

| Before | After |
| - | - |
|  |  |


nolanlawson:
It says Circle CI is having a partial outage, which may be the reason the tests weren't run. In any case this is unlikely to break tests. :slightly_smiling_face: 

And yes, looks great! Thanks :tada:

PR #2247, opened 2022-11-25
Ensure UI is readable by improving the contrast for all themes

NickColley:
An initial pass to improve overall contrast for themes, this does not fix all issues but trying not to do too much at once...

## Icons

Icons now have hover and active states on their touch area background rather than the icon themselves, this also simplifies the logic allowing both pressed and non-pressed state to share more states.

## Navigation
The background for the navigation when selected or interacted with have a better contrast, to ensure that it's clear what you're interacting with I have increased the width of the indicator line on desktop so we are relying on that being clear rather than the background itself.

## Loading indicators
Loading indicator and similar SVG filled icons now use the text colour to ensure good contrast

### DIFF
--- a/src/routes/_components/IconButton.html
+++ b/src/routes/_components/IconButton.html
@@ -52,21 +52,16 @@
    * regular styles
    */
 
-  :global(.icon-button:hover .icon-button-svg) {
-    fill: var(--action-button-fill-color-hover);
+  .icon-button:hover {
+    background: var(--action-button-bg-color-hover);
   }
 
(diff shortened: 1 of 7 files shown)


NickColley:
Getting close to this feeling ready for review, will mull it over another day and then get some screenshots but people are welcome to check out the deployment in the meantime and have a nosy.

NickColley:
To make it easier for us to review (and me to do screenshots!) I'm going to break this into smaller pull requests

PR #2254, opened 2022-11-25
fix: hide invisible content consistently

NickColley:
Some other parts of the interface for example URLs in profiles use the invisible class, move this to the top level global file so it'll be applied everywhere.

| Before | After |
| - | - |
|  |  |


nolanlawson:
Good catch!

nolanlawson (approved):

PR #2256, opened 2022-11-25
test: run tests on Mastodon v4

nolanlawson:


nolanlawson:
> Since Mastodon 3.6, filters can contain multiple keywords and are matched server-side.

[This](https://docs.joinmastodon.org/methods/filters/) is probably why the filter tests are busted – I'm doing client-side filtering. Not sure how to support Mastodon v3.6+ plus older versions and other servers like Pleroma and Misskey. Might need some feature detection that falls back to client-side filtering.

nolanlawson:
OK so apparently the test failure is caused by a true Mastodon v4 bug: https://github.com/mastodon/mastodon/issues/21965

Probably the long-term fix is #2265. For now we can just disable the test.

PR #2260, opened 2022-11-26
Make click on reposter's small avatar image go to reposter's account page

ringtailsoftware:
This change makes clicking on the small sized avatar go to the account page for the reposter, previously small and large images both linked to the original poster's account page

### DIFF
--- a/src/routes/_components/status/StatusHeader.html
+++ b/src/routes/_components/status/StatusHeader.html
@@ -1,6 +1,13 @@
 <div class="status-header {isStatusInNotification ? 'status-in-notification' : ''} {notificationType === 'follow' ? 'header-is-follow' : ''}">
   <div class="status-header-avatar {timelineType === 'pinned' || notificationType === 'poll' ? 'hidden' : ''}">
-    <Avatar {account} size="extra-small"/>
+    <a id={elementId}
+       href="/accounts/{accountId}"
+       rel="prefetch"
+       class="status-header-author"
+       title="{'@' + account.acct}"
+    >
+      <Avatar {account} size="extra-small"/>
+    </a>
   </div>
   <SvgIcon className="status-header-svg" href={icon} />
 
--- a/src/routes/_components/status/StatusHeader.html
+++ b/src/routes/_components/status/StatusHeader.html
@@ -3,7 +3,8 @@
     <a id={elementId}
        href="/accounts/{accountId}"
        rel="prefetch"
-       title="{'@' + account.acct}"
+       aria-hidden="true"
+       tabindex="-1"
     >
       <Avatar {account} size="extra-small"/>
     </a>
--- a/src/routes/_components/status/StatusHeader.html
+++ b/src/routes/_components/status/StatusHeader.html
@@ -3,7 +3,6 @@
     <a id={elementId}
        href="/accounts/{accountId}"
        rel="prefetch"
-       class="status-header-author"
        title="{'@' + account.acct}"
     >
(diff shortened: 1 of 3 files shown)


src/routes/_components/status/StatusHeader.html
```diff
@@ -1,6 +1,12 @@
 <div class="status-header {isStatusInNotification ? 'status-in-notification' : ''} {notificationType === 'follow' ? 'header-is-follow' : ''}">
   <div class="status-header-avatar {timelineType === 'pinned' || notificationType === 'poll' ? 'hidden' : ''}">
-    <Avatar {account} size="extra-small"/>
+    <a id={elementId}
```
NickColley, line 3: So right now this link is duplicating the text based 'Person has boosted' link. This is good for pointer users e.g. touch or mouse but for screen reader users it's duplicate.

Using 'title' to give the link a name is not sufficient for screen readers.

My recommendation here is to intentionally remove this duplicate link from being focusable with a keyboard or discoverable by a screen reader.

```html
<a id="..." href="..." aria-hidden="true" tabindex="-1">
   <img class="..." aria-hidden="true" alt="" title="" src="...">
</a>
```

NickColley (changes requested):

NickColley:
Love this proposal I've tried to click this myself, I have some blocking feedback related to accessibility that if we can resolve I think this'll be good to go.

ringtailsoftware:
Thanks @NickColley. I've had a play with VoiceOver and I think this is right?

NickColley:
Looks great, that aligns with how the main avatar works. Will let @nolanlawson give a final review thank you 👍

MarcoZehe:
The question is why this link has to be duplicated anyway and then be removed from keyboard and screen reader users afterwards? Is there no better way to construct the markup that doesn't require this horrible hack?

NickColley:
@MarcoZehe because the image is in a odd position relative to the link if we did it another way for example expanding the original links hit area we'd run into other issues I suspect so this is preferable. I've used this pattern many times and don't see it as a hack myself but always welcome to better solutions.

nolanlawson:
Yeah that is the solution I already took for the avatar image in general. It's duplicated by the account name, so I considered it pointless to keep it in the tab order or to make it visible for screenreaders. I wrote about my thought process [in this post](https://nolanlawson.com/2019/11/05/what-ive-learned-about-accessibility-in-spas/) (under "Subtleties with focus management").

The solution @NickColley suggests in https://github.com/nolanlawson/pinafore/pull/2260#discussion_r1032791676 seems to align with how I've currently handled things.

nolanlawson:
Also FWIW looks like one of the tests is broken due to a selector changing.

ringtailsoftware:
Thanks, I see the broken test in 010-focus.js
I guess the problem is about the normal sized avatar image no longer being the first `a`.
I don't understand enough to fix it without experimenting, so will need to set up a full dev environment with a Mastodon instance. That may take me a while.

nolanlawson:
Yeah running the tests locally is a beast! I need to learn how Docker works or something.

I took a look, and there were two things:

1. Changed the selector in the test
2. There was a true focus bug here – if you click the avatar, then press the back button, the focus is on the author name rather than the avatar image. That's because the IDs were the same. I made the IDs different

This _should_ fix the test; let's let CI figure it out ^

NickColley:
@nolanlawson great catch

NickColley (approved):

PR #2266, opened 2022-11-27
fix: improve button a11y for theme Sorcery

mlcdf:
Similar to https://github.com/tootcafe/mastodon/pull/183

Contract ratio is now > 7:1: https://contrast-ratio.com/#%23ae91e8-on-%23141414


nolanlawson:
Awesome, thank you!! :slightly_smiling_face:

PR #2268, opened 2022-11-28
feat: Add "a" keyboard shortcut to bookmark a toot.

### LINKED ISSUE #2237: Add a shortcut for the Bookmark action
The official Mastodon interface includes a Bookmark button next to every post, making it relatively easy to access. Likewise some mobile apps like Metatext offer "Bookmark" action, which again makes it very fast to do. In contrast, to bookmark a post in Pinafore you need to find it in the more menu which takes a lot of keystrokes. This is in contrast to the favorite command which is just 1 shortcut.

To this end I propose that the Bookmark command also get a keyboard shortcut. Since B, O, M, R and K are all taken I suggest the letter A. Like the Boost and Favorite actions it should also get a spoken confirmation through Aria if it's pressed.

jcsteh:
Fixes #2237.

### DIFF
--- a/src/intl/en-US.js
+++ b/src/intl/en-US.js
@@ -153,6 +153,7 @@ export default {
     <li><kbd>f</kbd> to favorite</li>
     <li><kbd>b</kbd> to boost</li>
     <li><kbd>r</kbd> to reply</li>
+    <li><kbd>a</kbd> to bookmark</li>
     <li><kbd>i</kbd> to open images, video, or audio</li>
     <li><kbd>y</kbd> to show or hide sensitive media</li>
     <li><kbd>m</kbd> to mention the author</li>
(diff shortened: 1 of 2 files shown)


nolanlawson:
Awesome, thank you! I just added one small thing: help text (for when you press the `?` key) and an aria live announcement of "bookmarked" or "unbookmarked."

jcsteh:
Note that when the toast appears, the alert will get read already, so I'm not sure the live region announcement is necessary.

jcsteh:
Sorry I forgot to update the help text. That was a silly oversight.

nolanlawson:
Ah yes, I forgot about the alert. The alert is actually preferable to the aria live region, so that sighted people get the feedback as well.

jcsteh:
In that case, are you happy for me to revert your live region change? The alert already gets displayed with my original code.

nolanlawson:
No worries, I reverted it. Now it just has the shortcut help text I added. Thanks a lot!

nolanlawson (approved):

PR #2269, opened 2022-11-28
feat: Allow image descriptions to be read automatically by screen readers without needing to expand media.

### LINKED ISSUE #2257: Allow image descriptions to be read automatically by screen readers without needing to expand media
When a screen reader user traverses the timeline, alt texts/image descriptions are not automatically read by their screen readers. Regardless of whether there is alt text associated with the media, the user must expand the media before they can discover if there is no description associated with it. Additional action should not be required to discover alt texxt or reading it.

There are a few potential solutions:

1. Indicate to the user that there is alt text associated with the media without passing on the alt text.
2. Indicate that there is alt text and provide a keyboard command to have the description read.
3. Indicate that there is alt text and automatically read the description. 

My personal preference is for either 2 or 3 with 3 being the best option. If the user wishes to look at the description in detail, there is always the option of expanding the media. That should be the last resort.

jcsteh:
Fixes #2257.

### DIFF
--- a/src/routes/_components/status/Status.html
+++ b/src/routes/_components/status/Status.html
@@ -273,6 +273,7 @@
       mediaAttachments: ({ originalStatus }) => (
         originalStatus.media_attachments
       ),
+      sensitiveShown: ({ $sensitivesShown, uuid }) => !!$sensitivesShown[uuid],
       sensitive: ({ originalStatus, $markMediaAsSensitive, $neverMarkMediaAsSensitive }) => (
         !$neverMarkMediaAsSensitive && ($markMediaAsSensitive || originalStatus.sensitive)
       ),
@@ -294,12 +295,12 @@
       ariaLabel: ({
         originalAccount, account, plainTextContent, shortInlineFormattedDate, spoilerText,
         showContent, reblog, notification, visibility, $omitEmojiInDisplayNames, $disableLongAriaLabels,
-        showMedia, sensitive, mediaAttachments, showPoll
+        showMedia, sensitive, sensitiveShown, mediaAttachments, showPoll
       }) => (
         getAccessibleLabelForStatus(originalAccount, account, plainTextContent,
           shortInlineFormattedDate, spoilerText, showContent,
           reblog, notification, visibility, $omitEmojiInDisplayNames, $disableLongAriaLabels,
-          showMedia, sensitive, mediaAttachments, showPoll
+          showMedia, sensitive, sensitiveShown, mediaAttachments, showPoll
         )
       ),
       showHeader: ({ notification, status, timelineType }) => (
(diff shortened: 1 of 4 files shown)


src/routes/_components/status/Status.html
```diff
@@ -270,6 +270,12 @@
         originalStatus.media_attachments &&
         originalStatus.media_attachments.length
       ),
+      mediaAttachments: ({ originalStatus }) => (
+        originalStatus.media_attachments
+      ),
+      sensitive: ({ originalStatus, $markMediaAsSensitive, $neverMarkMediaAsSensitive }) => (
+        !$neverMarkMediaAsSensitive && ($markMediaAsSensitive || originalStatus.sensitive)
```
jcsteh, line 277: It's not great that this duplicates code in StatusMediaAttachments, but I wasn't sure how I could get that info without some serious re-plumbing and I'm not quite experienced enough with this code base to pull that off.

jcsteh:
Oops. I broke it. I'll look into it later.

src/routes/_components/status/Status.html
nolanlawson: One way to avoid the duplication would be to pass `sensitive` in to `StatusMediaAttachments`. But It's not a big deal either way.

tests/spec/120-status-aria-label.js
```diff
@@ -11,7 +11,7 @@ test('aria-labels for statuses with no content text', async t => {
   await t
     .hover(getNthStatus(1))
     .expect(getNthStatus(1).getAttribute('aria-label')).match(
-      /foobar, has media, (.+ ago|just now), @foobar, Public/i
+      /foobar, has media, kitteh, (.+ ago|just now), @foobar, Public/i
```
nolanlawson, line 14: Thank you for fixing the test! :heart:

src/routes/_components/status/Status.html
nolanlawson: So unfortunately it looks like there's an issue here. If there is a sensitive media, and you click the "show sensitive media" button, then the aria-label doesn't change. It permanently shows the status in "sensitive" mode.

Maybe this wouldn't matter much in practice, but it would matter if you e.g. are scrolling through the timeline, press the `y` button to show the sensitive media, and the aria-label doesn't change at all.

I think in order to get this to work, we would have to plumb from the `StatusMediaAttachments` back to the `Status`. It looks like I already have a store value called `sensitivesShown` that does this. I can look into this.

https://github.com/nolanlawson/pinafore/blob/a775bd9193a8d3b0f24201d09667943210a3e1f4/src/routes/_components/status/StatusMediaAttachments.html#L217-L237

nolanlawson: This should do it! 470cb15d

nolanlawson (approved):

Issue #2249, opened 2022-11-25
Notification page has low contrast for toots


NickColley:
![](https://user-images.githubusercontent.com/2445413/203989852-95f0883e-f67f-4d49-bde7-b6cf78ab745a.png)

 think perhaps a way forwards is to remove the status toolbar controls from these toots and increase the contrast so it at least meets WCAG requirements.

![](https://user-images.githubusercontent.com/2445413/203990463-89d6a618-d660-4807-85e0-1a729629afe5.png)

nolanlawson:
I guess these ones don't technically need the toolbar, yeah. You can just click through to see the controls. Maybe that's fine anyway, since the UI is probably already cluttered with them due to no coalescing: #2184 

Increasing the contrast on these is great IMO as long as there is still some affordance that this is a "status within a notification" and in some sense "inert." Also the "so-and-so favorited" text should stand out more than the toot text itself, since that's what's really important. It's a tough one. :confused:

PR #2278, opened 2022-11-29
Improve notification page readability

NickColley:
Use lowest possible contrast gray that meets WCAG requirements for very deemphasised text.

Makes the notification page more readable without compromising access as much.

Closes https://github.com/nolanlawson/pinafore/issues/2249

### DIFF
--- a/src/routes/_components/IconButton.html
+++ b/src/routes/_components/IconButton.html
@@ -52,25 +52,16 @@
    * regular styles
    */
 
-  :global(.icon-button:hover .icon-button-svg) {
-    fill: var(--action-button-fill-color-hover);
+  .icon-button:hover {
+    background: var(--action-button-bg-color-hover);
   }
 
-  :global(.icon-button.not-pressable:active .icon-button-svg,
-  .icon-button.same-pressed:active .icon-button-svg) {
-    fill: var(--action-button-fill-color-active);
+  .icon-button:active {
+    background: var(--action-button-bg-color-active);
   }
 
   :global(.icon-button.pressed.not-same-pressed .icon-button-svg) {
     fill: var(--action-button-fill-color-pressed);
-  }
-
-  :global(.icon-button.pressed.not-same-pressed:hover .icon-button-svg) {
-    fill: var(--action-button-fill-color-pressed-hover);
-  }
-
-  :global(.icon-button.pressed.not-same-pressed:active .icon-button-svg) {
-    fill: var(--action-button-fill-color-pressed-active);
   }
 
   /*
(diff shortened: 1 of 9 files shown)


src/routes/_components/status/StatusToolbar.html
```diff
@@ -67,6 +67,11 @@
       height: 20px;
     }
   }
+
+  /* Hack, do this properly with state... */
```
NickColley, line 71: Just hacked this so people could see how it looks...

NickColley:
@nolanlawson this one is for review as well really but I'd need to do the work to properly pass data through to hide the status toolbar rather than hacking it if you're into this.

Alternatively we just keep the status toolbar and keep the colour changes, which I think meet the requirement of being deemphasised but still readable.

src/routes/_components/status/StatusToolbar.html
nolanlawson: No worries, I fixed it to use state: 6fcdb149

I can't recall why I made `isStatusInNotification` false if the notification is a mention. I think it was to give a different style to mentions vs other kinds of notifications. In any case I just added `isStatusInNotificationIncludingMentions` and it works. :man_shrugging:

src/routes/_components/IconButton.html
```diff
@@ -52,27 +52,18 @@
...
+  .icon-button:hover {
+    background: var(--action-button-bg-color-hover);
   }
 
-  :global(.icon-button.not-pressable:active .icon-button-svg,
-  .icon-button.same-pressed:active .icon-button-svg) {
-    fill: var(--action-button-fill-color-active);
+  .icon-button:active {
+    background: var(--action-button-bg-color-active);
   }
```
nolanlawson, line 61: IIUC the goal here is to use background colors rather than the colors of the SVG buttons to communicate hover vs active states?

Personally I liked the subtlety of the previous design here, especially for the boost/fav icons in the status toolbar. I think it meshes well with the animation when you press it too. Only one thing changes in the process of hover-click-unclick: the color of the icon.

Given that we now have the checkmark for the boost/fav icons, is this a must-have?

nolanlawson (changes requested):
Thanks for this PR! We're getting closer to fixing contrast issues. :slightly_smiling_face: 

I think getting rid of the `StatusToolbar` on the notifications page is a great idea. It just adds visual noise anyway. And without it, we can avoid the grayed-out look for the icons.

For the hover/active states of the toolbar icons elsewhere, I'm more a fan of the old design rather than changing the background color on the icons. I left a comment about that.

Thanks again!

NickColley:
I missed up a rebase this is not dependent on the other work so will fix that and should be easier to review then

NickColley:
Continued in #2302

PR #2291, opened 2022-12-03
fix: fix nav animations

### LINKED ISSUE #2290: Indicator doesn't animate correctly during page transitions
Unfortunately #2274 seems to have regressed the "indicator" animation. Instead of smoothly animating the white bar over a dark background, the background jumps from dark to light, and then the white bar animates "into" it.

Old Pinafore on the left, new Pinafore on the right (animation slowed down from 0.3s to 2s for clarity):

https://user-images.githubusercontent.com/283842/205456145-06ec459a-acb9-47d3-91df-116ce6e3f897.mp4

The easiest way to repro this on desktop is to press the 1-6 keys to navigate without hovering. It also repros on Firefox on Android for me, by touching the nav buttons.

I played around with it a bit, but it's really hard to get this to work correctly. The main reason it was working before is that that "indicator wrapper" (i.e. background) did _not_ have a different color in the hover or active state. So it was easy to animate the indicator bar over it.

There is probably a way to do this correctly with the new hover/active states, but I haven't figured it out yet.

nolanlawson:
Fixes #2290.

https://user-images.githubusercontent.com/283842/205460527-0c795d55-436b-49c2-8606-b2821ff7cbd7.mp4

My solution was a few things:

1. Instead of using colors with transparency (which I found impossible to animate correctly due to backgrounds messing with the foreground color), use SASS `mix()` to achieve basically the same colors. (I eyeballed it, so not super scientific, but it seems close.)
2. When animating the indicator, mutate the color of the wrapper around (behind) it
3. Refactor the CSS vars to make the code easier to understand (`--nav-indicator-bg`, `--nav-indicator-bg-active`, `--nav-indicator-bg-hover`).

I tested 3 methods: using the keyboard to change the page, using the mouse on desktop, and simulating touch on a mobile phone. They all look pretty good to me.

FYI @NickColley

### DIFF
--- a/src/routes/_components/NavItem.html
+++ b/src/routes/_components/NavItem.html
@@ -12,8 +12,8 @@
     />
     <span class="nav-link-label">{label}</span>
   </div>
-  <div class="nav-indicator-wrapper">
-    <div class="nav-indicator" ref:indicator></div>
+  <div class="nav-indicator-wrapper {animationClass}" ref:wrapper>
+    <div class="nav-indicator {animationClass}" ref:indicator></div>
   </div>
 </a>
 <style>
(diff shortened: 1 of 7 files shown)


NickColley:
Looking very nice 👍

PR #2296, opened 2022-12-04
Fix rich push notifications for single-instance situations

tbroyer:


nolanlawson:
@tbroyer Thanks for opening this PR! Could you provide some more detail on what bug this fixes exactly? I don't use push notifications myself, so I'm not super familiar with this part of the code. Thanks!

nolanlawson:
Ah I see, I missed https://github.com/nolanlawson/pinafore/issues/1663. I will study to see what is going on here.

nolanlawson:
I'm confused about what's going on here and what this fixes. Maybe my brain is dead because it's Sunday morning, but I just don't get #1663 at all. https://github.com/nolanlawson/pinafore/issues/1663#issuecomment-1336510335

src/service-worker.js
```diff
@@ -185,8 +193,10 @@ self.addEventListener('push', event => {
 
 async function showSimpleNotification (data) {
   await self.registration.showNotification(data.title, {
+    badge: '/icon-push-badge.png',
     icon: data.icon,
     body: data.body,
+    tag: data.notification_id,
```
nolanlawson, line 199: If this is unrelated to this PR, could you open it as a separate PR? Thanks!

src/service-worker.js
```diff
@@ -169,8 +171,14 @@ self.addEventListener('fetch', event => {
 self.addEventListener('push', event => {
   event.waitUntil((async () => {
     const data = event.data.json()
-    const { origin } = event.target
+    const knownInstances = await getKnownInstances()
+    if (knownInstances.length !== 1) {
+      // TODO: find a way to determine the pushing instance from the PushEvent
+      await showSimpleNotification(data)
```
nolanlawson, line 177: It looks like we are eagerly showing a simple notification rather than a rich one, whereas the original logic was to show a rich one and fallback to a simple one. Why would we prefer to always show a simple notification rather than a rich one, when there is only one instance in the known instances list?

tbroyer: More details in https://github.com/nolanlawson/pinafore/issues/1663#issuecomment-1336548657

The problem is that there's no way to compute the `origin` from the `event` alone.

src/service-worker.js
tbroyer: Done #2299

src/service-worker.js
```diff
@@ -169,8 +171,14 @@ self.addEventListener('fetch', event => {
...
+      return
+    }
 
+    const origin = basename(knownInstances[0])
```
nolanlawson, line 181: This is an unfortunate solution because a user might log into 2 instances and enable notifications on the second one.

It seems like a better system would be for us to log somewhere _which_ instance they've enabled notifications for. I.e. we can enforce the "you can only have notifications for one instance at once time" system ourselves.

tbroyer: Indeed it's not ideal. For the case where a user logged into 2 instances, they'd have "simple" notifications even if they did enable notifications only on one of them.

Note that this line gets the *single* element from `knownInstances`, as the case where there are more than one (more accurately, not exactly one) is handled above; so it's not a matter of "enabling notifications on the second one".

src/service-worker.js
tbroyer: The idea here is that:
* if we only know about a single instance, the notification should come from it so we use it (line 181) and show a rich notification
* otherwise, we have no way to determine the `origin`, so fallback to a simple notification.

Pending deeper changes on the way enabling notifications is handled (e.g. only allow enabling notifications on a single instance) or changes on Mastodon's side (including the URL to the notification, or at least the instance, in the PushEvent), this PR only fixes things in the case where the user logged in to a single instance. Not ideal but that's the simplest change to at least get rich notifications for some users (most of them? a good proportion of them? I'm certainly one of them).

src/service-worker.js
```diff
@@ -169,8 +171,14 @@ self.addEventListener('fetch', event => {
 self.addEventListener('push', event => {
   event.waitUntil((async () => {
     const data = event.data.json()
-    const { origin } = event.target
+    const knownInstances = await getKnownInstances()
```
nolanlawson, line 174:
```suggestion
    // If there is only once instance, then we know for sure that the push notification came from it
    const knownInstances = await getKnownInstances()
```

src/service-worker.js
```diff
@@ -169,8 +171,15 @@ self.addEventListener('fetch', event => {
 self.addEventListener('push', event => {
   event.waitUntil((async () => {
     const data = event.data.json()
-    const { origin } = event.target
+    // If there is only once instance, then we know for sure that the push notification came from it
+    const knownInstances = await getKnownInstances()
+    if (knownInstances.length !== 1) {
+      // TODO: find a way to determine the pushing instance from the PushEvent
```
nolanlawson, line 177:
```suggestion
      // TODO: Mastodon currently does not tell us which instance the push notification came from.
      // So we have to guess and currently just choose the first one. We _could_ locally store the instance that
      // currently has push notifications enabled, but this would only work for one instance at a time.
      // See: https://github.com/mastodon/mastodon/issues/22183
```

nolanlawson (approved):
This is better than the status quo, even if it's not ideal.

Issue #2293, opened 2022-12-04
Retweet button is smaller than others - `fa-retweet` is not a square


xnuk:
![](https://user-images.githubusercontent.com/3071003/205494397-e47570c9-4b52-4188-b3b1-3143b54d22bd.png)

Size of the retweet button is 2048x1792:

https://github.com/nolanlawson/pinafore/blob/380d2a0d451cd6d74483a44fb5ef5f4a263c728b/src/thirdparty/font-awesome-svg-png/white/svg/retweet.svg#L1

While sizes of all `IconButton`s are 24x24px (or smaller square sizes):

https://github.com/nolanlawson/pinafore/blob/380d2a0d451cd6d74483a44fb5ef5f4a263c728b/src/routes/_components/IconButton.html#L31-L36

That makes bit smaller than other icons.

nolanlawson:
thanks for opening the issue! This looks like a duplicate of #916.

Currently I would prefer not to do hacks just for a particular icon; I would rather fix the source svg.

Issue #916, opened 2019-01-26
Proposal: Improving status button icons


charlag:
Hey!
I've always felt that something is off with "boost" icon and now I think I know the reason: it's smaller.

![screenshot from 2019-01-26 14-47-32](https://user-images.githubusercontent.com/3099142/51788131-ded8fd00-217a-11e9-841b-86cb5352585b.png)

All icons are of the same width but because boost icon has different proportions it always seems smaller to me.
I would propose to increase the width of the boost icon to compensate it's aspect ratio:
![screenshot from 2019-01-26 15-04-43](https://user-images.githubusercontent.com/3099142/51788229-daf9aa80-217b-11e9-9124-dbcadcd0cc29.png)

I would also propose some non-color identification of boost/fav status. With star it's simple: empty and filled star. Boost icon would require much more creativity and I have no good ideas, maybe other folks do.

nolanlawson:
Thanks for reporting. You can hover over the icons to see a tooltip showing what the state is; this tooltip is the same as the aria label that's provided to screenreaders. Empty star would be tricky because I believe this is not provided by font awesome, plus it would conflict with the current animation.

As for the boost button yes, fontawesome arguably should make it wider to match the other icons. This seems like a straightfoward fix assuming everything still lines up correctly.

charlag:
I just want to point out that they *made* it wider, in fact it's *too wide* and that's why it doesn't match. You can make the same change in the CSS as I did to compensate it (or I could try to send a PR if needed)

Heartade:
Weird, but the icon is not a square even in the latest release of FontAwesome. Mind if I edit the `retweet.svg` manually in the `static` directory and submit a PR?

PR #2297, opened 2022-12-04
fix: redesigned boost icon to fix alignment (#916)

Heartade:
Tried to fix the alignment and viewport size of the original FontAwesome icon.
The original FA icon svg is a little bit biased to the bottom, and has `2048x1792` viewport size, causing it to be smaller compared to other icons with `1792x1792` viewport size when crammed into `.icon-button-svg`'s `24x24` size.

I made this icon from scratch, but tried to make it look as natural as possible when used with other FA icons.

image
_the original icon_

image
_the new icon_

The boost icon still looks smaller than others (#2293), due to the design being wider and still having to fit in a square.

Also considering it is original svg, not part of the FA set, we should probably consider moving it to something like `/static/reblog.svg`, not `/thirdparty/font-awesome-svg-png/white/svg/retweet.svg`. That would require altering `svgs.js`, so I'm not going that far in this PR.


nolanlawson:
I don't see any problem with leaving the SVG file where it is. I modified plenty of the thirdparty packages that I vendor'ed into the project – the important thing IMO is just to keep the original license.

Thanks for your work on this BTW! It looks amazing.

Just to be safe, I checked, and the `icons.svg` file is also still roughly the same file size (24.9k -> 25k). Marvelous work.

nolanlawson (approved):

NickColley (approved):

NickColley (commented):
image

When the icon is used in a smaller context it the detail feels lost is there a way to increase the height without making the 'arms' of the icon so thick?

PR #2298, opened 2022-12-04
Intl/ko

Heartade:
Add Korean translation for Pinafore. (Work in progress)


nolanlawson:
Thanks for opening this PR! Just so you're aware, Pinafore currently only supports translations if you host your own copy of it. https://github.com/nolanlawson/pinafore/blob/master/docs/Internationalization.md

Heartade:
> Thanks for opening this PR! Just so you're aware, Pinafore currently only supports translations if you host your own copy of it. https://github.com/nolanlawson/pinafore/blob/master/docs/Internationalization.md

I actually found it to work only when I manually edit `bin/getIntl.js` file in addition to providing `LOCALE` environment variable. Is this expected behavior, or should I file a bug?

Also, hosting multiple builds with different locales within the same host e.g. `en-us.pinafore.social` could be a way to support multiple locales while dynamic internationalization is yet to arrive.

nolanlawson:
> I actually found it to work only when I manually edit bin/getIntl.js file in addition to providing LOCALE environment variable. Is this expected behavior, or should I file a bug?

No, that sounds about right, that's how it's supposed to work. See: https://github.com/nolanlawson/pinafore/blob/master/docs/Internationalization.md

> Also, hosting multiple builds with different locales within the same host e.g. en-us.pinafore.social could be a way to support multiple locales while dynamic internationalization is yet to arrive.

Yes, this is what I had in mind, but didn't get around to it.

Issue #2267, opened 2022-11-28
Notify screen reader users when character count is at or near 500


nolanlawson:
Got a suggestion about this: https://mas.to/@norrumar/109411900076228505

The character count is currently accessible, but it doesn't notify you in real time when you're getting close to 500. Arguably we could use the `aria-live=polite` region for this.

nvdaes:
Thanks @nolanlawson 
I reported this in the mentioned toot. My username in Mastodon is norrumar@mas.to, andmy website at nvdaes.github.io is verified in my profile, to check that nvdaes corresponds to my profile in Mastodon.
I think this maybe reported, for example, at 495 characters, and later you may decide to create an option in general settings to save the preferred limit in localStorage or something.
I've cloned this repo and just installed yarn for testing or contribute if needed.

nvdaes:
cc: @mhameed

NickColley:
Just technical thing, not all instances have a limit to 500 so this should be based on the instance's upper limit.

nvdaes:
@NickColley , I agree. I have look at source code and I thought about the following approach, but I have problems running yarn run dev and I cannot test and so I don't know if this is right:

in routes/components/lengthIndicator.html, in the span element,

aria-live={lengthVerbosity} aria-atomic="true"

And in the script

let lengthVerbosity = (overLimit || lengthToDisplayDeferred < 16) ? "polite" : "off"

nvdaes:
I've tested setting aria-live attribute in lengthIndicator.html component to polite and assertive, but seems not to work as expected by me. NVDA doesn't notify changes though aria-label changes when I type characters in the compose edit box.
I don't know if this is related to the fact that the span text contents is not changing, just aria-label, or due to something not considered by me.

nvdaes:
I may create a pull request today that fixes this issue for me.
My approach is the following:

1. Remove aria-label.
2. When lengthToDisplayDeferred is minor than 10, lengthVerbosity variable is set to polite, else to off.
3. Screen readers are notified just to changes in the real contents of the span, without addin extra info like "characters remaining" or "overlimit", since we can listen or read in our braille displays the minus symbol before numbers, and at least for me this is less verbose, and of course easierto implement.
4. I've removed an import not used, according to the linter advice.

I have tested with NVDA screenreader, typing and deleting characters, with speech and braille, without visual feedback.
In case you disagree with this implementation, please letme know.

nolanlawson:
[...]

PR #2300, opened 2022-12-05
Report remaining chars

nvdaes:


nvdaes:
Fixes issue #2267

src/routes/_components/LengthIndicator.html
```diff
@@ -42,11 +41,11 @@
...
       lengthToDisplay: ({ length, max }) => max - length,
-      lengthLabel: ({ overLimit, lengthToDisplayDeferred }) => {
-        if (overLimit) {
-          return formatIntl('intl.overLimit', { count: -lengthToDisplayDeferred })
+      lengthVerbosity: ({ lengthToDisplayDeferred }) => {
+        if (lengthToDisplayDeferred > 10) {
+          return 'off'
         } else {
-          return formatIntl('intl.underLimit', { count: lengthToDisplayDeferred })
+          return 'polite'
```
nolanlawson, line 48: Now that `intl.overLimit` and `intl.underLimit` are unused, we can remove them from `en-US.js` (and the other intl files in the same directory).

src/routes/_components/LengthIndicator.html
```diff
@@ -42,11 +41,11 @@
     store: () => store,
     computed: {
       lengthToDisplay: ({ length, max }) => max - length,
-      lengthLabel: ({ overLimit, lengthToDisplayDeferred }) => {
-        if (overLimit) {
-          return formatIntl('intl.overLimit', { count: -lengthToDisplayDeferred })
+      lengthVerbosity: ({ lengthToDisplayDeferred }) => {
```
nolanlawson, line 44:
```suggestion
      lengthVerbosity: ({ lengthToDisplayDeferred }) => {
        // When approaching the limit, notify screen reader users
```

nolanlawson (approved):

src/routes/_components/LengthIndicator.html
```diff
@@ -42,11 +44,12 @@
...
+        // When approaching the limit, notify screen reader users
+        if (lengthToDisplayDeferred > WARN_THRESHOLD) {
```
nolanlawson, line 49:
```suggestion
        if (lengthToDisplayDeferred > WARN_THRESHOLD || lengthToDisplayDeferred <= 0) {
```

@nvdaes It seems to me that we should keep warning if the user reaches 0 or goes beyond the limit (-1, -2, etc.). Let me know what you think of this change. :arrow_up:

nolanlawson (changes requested):
See question above, otherwise LGTM

nolanlawson (approved):

nvdaes:
Thanks @nolanlawson. Several times I had to rewrite a toot since I exceeded the limit of characters and this may help a lot when we type a long message

Issue #1284, opened 2019-06-18
Official docker images?


pwFoo:
Hi,
is a official docker image available? Would be nice if an image is built with each new release and pushed to docker hub.

nolanlawson:
I don't plan on publishing an official Docker image right now. I don't know enough about Docker to support this, plus I have an inherent uneasiness toward any central repository (hence why Pinafore is not on the App Store or Play Store; GitHub is an exception). Thanks for asking! :)

PR #2301, opened 2022-12-05
Add Docker image workflow

### LINKED ISSUE #2148: dockerhub official builds please!
it'd be really good to have official builds of this in dockerhub from this repo, so that folks can use them with container orchestration systems without having to build and host the containers themselves. it's fairly difficult for me to make my own docker build and publish job whenever your upstream codebase does a release, but it'd be pretty easy for you to do that, and then folks would know the image came from you.

samsapti:
Added a GH Actions workflow to automatically publish Docker images with the following tags:
- SemVer tags (`X.Y.Z`, `X.Y`, `X`)
- `latest` tag
- `edge` tag that follows the master branch

Closes #2148. Regarding #1284, if you don't want to use Docker Hub, GHCR could be an alternative, since the project is already on GitHub. Having an official Docker image somewhere significantly eases self-hosting.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -1,0 +1,42 @@
+name: Build and push Docker image
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+  pull_request:
+    paths:
+      - .github/workflows/build-image.yml
+      - Dockerfile
+jobs:
+  build-image:
+    name: Build and push Docker image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Extract metadata for Docker image
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ secrets.DOCKERHUB_USERNAME }}/pinafore
+          flavor: |
+            latest=auto
+          tags: |
+            type=semver,pattern={{version}}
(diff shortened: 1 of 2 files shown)


nolanlawson:
Thanks for the PR! I have zero knowledge about Docker, so bear with me. :slightly_smiling_face: 

Questions:

1. I suppose I need to create a Docker account and set the env vars for the username/password?
2. This will run for every commit and tag, as I understand it. I'm a bit concerned this will push me over the limit of my available GitHub Actions budget, which is already being eaten up by the CI tests. But maybe I can turn it off if I start approaching the limit.
3. How do I advertise this? In the readme? What instructions do I give people? Do I have to create some kind of profile page on Docker itself?

samsapti:
@nolanlawson no problem at all! I'll try to answer your questions:

1. Yes, that's correct. You need to create an account and put your credentials in the GH repo secrets.
2. You can certainly turn it off if you want. Alternatively, we can make it so it only creates a Docker image on git tags / releases (essentially removing the `edge` image tag, which is the one created on every commit).
3. I would write it in the README somewhere, and edit your existing `docker-compose.yml` file to use the official image. Usually people use the README from GH on the Docker image page over on Docker Hub (I actually don't know whether that's a built-in feature on Docker Hub or not).

nolanlawson:
Thanks for the details. It will probably take me quite some time to figure out enough about Docker to set up the account, put the proper instructions in the README, and make the proper modifications to the existing Dockerfile. I can't promise I'll get to this soon.

samsapti:
@nolanlawson you shouldn't need to modify the Dockerfile. You only need to make an account on [Docker Hub](https://hub.docker.com), and then create a repo named `pinafore`. Assuming you'll be using the same username as on GH, I've modified the `docker-compose.yml` file for you.

I can also help out with the README if you want me too.

Aeolun:
I came here to make this only to see that someone already did. Thanks @samsapti !

PR #2302, opened 2022-12-05
Improve notifications page readability

NickColley:
Use lowest possible contrast gray that meets WCAG requirements for very deemphasised text.

Removes the status toolbar when its a notification for your own toot

Makes the notification page more readable without compromising access as much.

Closes https://github.com/nolanlawson/pinafore/issues/2249

### Light

| Before | After |
| - | - |
| notifications page with hard to read text in toot notification | notifications page with clearer text and no buttons in toot notification |

### Light

| Before | After |
| - | - |
| notifications page with hard to read text in toot notification | notifications page with clearer text and no buttons in toot notification |

### DIFF
--- a/src/scss/themes/_base.scss
+++ b/src/scss/themes/_base.scss
@@ -81,8 +81,8 @@
   --deemphasized-text-color: #{$deemphasized-color};
   --focus-outline: #{$focus-outline};
 
-  --very-deemphasized-link-color: #{rgba($anchor-color, 0.6)};
-  --very-deemphasized-text-color: #{rgba(#666, 0.6)};
+  --very-deemphasized-text-color: #757575;
+  --very-deemphasized-link-color: var(--very-deemphasized-text-color);
 
   --status-direct-background: #{darken($body-bg-color, 5%)};
   --main-theme-color: #{$main-theme-color};
--- a/src/scss/themes/_dark.scss
+++ b/src/scss/themes/_dark.scss
@@ -1,5 +1,5 @@
 :root {
-  $deemphasized-color: lighten($main-bg-color, 45%);
+  $deemphasized-color: lighten($main-bg-color, 54%);
 
   --action-button-deemphasized-fill-color: #{$deemphasized-color};
   --action-button-deemphasized-fill-color-hover: #{lighten($deemphasized-color, 22%)};
@@ -12,8 +12,8 @@
 
   --deemphasized-text-color: #{$deemphasized-color};
 
-  --very-deemphasized-link-color: #{rgba($anchor-color, 0.8)};
-  --very-deemphasized-text-color: #{lighten($main-bg-color, 32%)};
+  --very-deemphasized-text-color: #{lighten($main-bg-color, 44%)};
+  --very-deemphasized-link-color: var(--very-deemphasized-text-color);
 
   --status-direct-background: #{darken($body-bg-color, 5%)};
   --main-theme-color: #{$main-theme-color};
--- a/tests/spec/005-status-types.js
+++ b/tests/spec/005-status-types.js
@@ -35,3 +35,10 @@ test('shows direct vs followers-only vs regular in notifications', async t => {
     .eql('Cannot be boosted because this is a direct message')
     .expect($(`${getNthStatusSelector(5)} .status-toolbar button:nth-child(2)`).hasAttribute('disabled')).ok()
 })
+
+test('hides status toolbar on notification page', async t => {
+  await loginAsFoobar(t)
+  await t
+    .navigateTo('/notifications')
+    .expect($(`${getNthStatusSelector(1)} .status-toolbar`).exists).notOk()
+})
--- a/tests/spec/026-shortcuts-notification.js
+++ b/tests/spec/026-shortcuts-notification.js
@@ -1,10 +1,10 @@
 import {
   closeDialogButton,
   composeModalInput,
-  getNthFavoritedLabel,
   getNthStatus,
   getUrl, modalDialog, notificationsNavButton,
-  isNthStatusActive, goBack
+  isNthStatusActive, goBack,
+  getNthFavoritedLabel
 } from '../utils'
 import { loginAsFoobar } from '../roles'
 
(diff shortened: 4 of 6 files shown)


NickColley:
Think legit CI failure will update tests soon

NickColley:
Hmm this also removes the toolbar for replies and toots like that, perhaps we want to keep them for that? Only remove them when it's your own toot...

tests/spec/005-status-types.js
```diff
@@ -18,20 +18,9 @@ test('shows followers-only vs regular in home timeline', async t => {
     .expect($(`${getNthStatusSelector(3)} .status-toolbar button:nth-child(2)`).hasAttribute('disabled')).ok()
 })
 
-test('shows direct vs followers-only vs regular in notifications', async t => {
+test('hides status toolbar on notification page', async t => {
```
NickColley, line 21: I need to figure out getting these test running locally and change it so we have a test that shows the status bar is missing on all notifications apart from mentions...

src/scss/themes/_base.scss
```diff
@@ -81,8 +81,8 @@
   --deemphasized-text-color: #{$deemphasized-color};
   --focus-outline: #{$focus-outline};
 
-  --very-deemphasized-link-color: #{rgba($anchor-color, 0.6)};
-  --very-deemphasized-text-color: #{rgba(#666, 0.6)};
+  --very-deemphasized-text-color: #757575;
+  --very-deemphasized-link-color: var(--very-deemphasized-link-color);
```
nolanlawson, line 85: Typically I try not to mix CSS vars and SASS vars in the `scss` files but I guess it doesn't really matter.

tests/spec/005-status-types.js
nolanlawson: Your current test looks good to me. In case it helps, I use a Podman container running Ubuntu 20.04 that basically runs the steps that the CI uses to install everything: 

https://github.com/nolanlawson/pinafore/blob/4a6907bbdc72659a86065ebc62afcc45e8aa2de2/.github/workflows/e2e-tests-readonly.yml#L46-L56

tests/spec/026-shortcuts-notification.js
```diff
@@ -11,23 +10,6 @@ import { loginAsFoobar } from '../roles'
...
-    .expect(getUrl()).contains('/notifications')
-    .expect(getNthStatus(1 + idx).exists).ok({ timeout: 30000 })
-    .expect(getNthFavoritedLabel(1 + idx)).eql('Favorite')
-    .pressKey('j '.repeat(idx + 1))
-    .expect(isNthStatusActive(1 + idx)()).ok()
-    .pressKey('f')
-    .expect(getNthFavoritedLabel(1 + idx)).eql('Unfavorite')
-    .pressKey('f')
-    .expect(getNthFavoritedLabel(1 + idx)).eql('Favorite')
-})
```
nolanlawson, line 29: Well that's kind of a shame that the shortcuts don't work anymore. But I guess it isn't the end of the world.

tests/spec/005-status-types.js
NickColley: Thanks appreciated! Will try this tomorrow

src/scss/themes/_base.scss
nolanlawson: Actually this is a circular reference. The CSS var is referencing itself. It looks like the browser treats this as a no-op.

Apparently this causes links inside of statuses in notifications (e.g. "so-and-so favorited your toot") to appear the same color as the rest of the text (e.g. `example.com` below):

Screenshot from 2022-12-10 11-06-08

tests/spec/026-shortcuts-notification.js
nolanlawson: I've added this test back in: fda15ac8 . We can just favorite another toot.

nolanlawson (changes requested):
Thanks a bunch, this largely LGTM, except for the circular var issue I mention above.

Sorry for taking so long to review, but I've had a bunch of stuff going on in my life. :confused: Also there are a deluge of PRs on Pinafore right now. But thanks again for your accessibility work!

NickColley:
Don't worry about the time to review, maybe I can help review and merge some stuff in future?

nolanlawson:
You have committer access, so you should be able to, yeah. :)

I will need to hand off this project at some time anyway if I'm being honest with myself: https://nolanlawson.com/2022/11/22/thoughts-on-mastodon/

Next step is probably to create a GitHub org for it.

NickColley:
Okay when I tried to get the tests running locally I accidentally upgraded my machine to the latest Ubuntu version and bricked it but now I've got a fresh install (that's how they get ya) so will hopefully finish this soon.

NickColley:
@nolanlawson this is ready to merge once you're happy with it!

nolanlawson:
> I accidentally upgraded my machine to the latest Ubuntu version and bricked it 

Sorry to hear about that. This is actually exactly why I finally learned to use Podman – it was too hard to get certain versions of Mastodon to work with my Ubuntu version.

FWIW, this is what I use:

```sh
podman run -dt -p 4002:4002 -p 3000:3000 -p 4000:4000 -p 10000:10000 -p 3035:3035 -v ~/workspace/pinafore:/mnt ubuntu:18.04
```

(Ubuntu 20.04 should also work.)

src/scss/themes/_dark.scss
```diff
@@ -12,8 +12,8 @@
 
   --deemphasized-text-color: #{$deemphasized-color};
 
-  --very-deemphasized-link-color: #{rgba($anchor-color, 0.8)};
-  --very-deemphasized-text-color: #{lighten($main-bg-color, 32%)};
+  --very-deemphasized-text-color: #{lighten($main-bg-color, 44%)};
+  --very-deemphasized-link-color: var(--very-deemphasized-text-color);
```
nolanlawson, line 16: `--very-deemphasized-link-color` is redundant now that it's the same as `--very-deemphasized-text-color`. And it's only used in one place:

https://github.com/nolanlawson/pinafore/blob/774aa7a21c706d62a8ac08510c933ebaf5376eca/src/routes/_components/status/StatusContent.html#L66-L68

The impact of this is that links only have a color that differentiates them from the surrounding text if they're in a mention notification rather than a fav/boost notification:

Screenshot from 2022-12-17 09-36-50

Technically this does not meet [WCAG link requirements](https://webaim.org/blog/wcag-2-0-and-link-colors/) since the only thing distinguishing the link is that they're underlined when you hover over them.

To be fair, though, the existing colors have the exact same problem:

Screenshot from 2022-12-17 09-42-45

Maybe it would be best to just use the same (blue) color as links in mention notifications? Or a slightly different shade of blue? We don't need to solve this in this PR, though, since it's a pre-existing issue.

nolanlawson (approved):

nolanlawson:
LGTM, I had a couple nitpicks, but if you want to resolve them in this PR please go ahead, or else feel free to click the merge button and we can do it in another PR. Thanks!

src/scss/themes/_dark.scss
NickColley: One approach might be to just turn links into plain text for these types of notifications and make it one big clickable area, but agreed existing issue that could be improved.

PR #2303, opened 2022-12-05
fix: correctly label toot edited notification

### LINKED ISSUE #2280: Support toot edited notification
When a toot is edited on a newer Mastodon or Pleroma server, everyone who boosted that toot gets a notification.

In Pinafore, currently, this notification shows up as just the toot, with no context for why it's showing up as a notification:

notif

The equivalent notification on Mastodon 4.0.2 web is labeled, "[person's name] edited a post":

notif2

(Hope this visual still makes sense with pixelization; I'm trying to preserve the tooter's privacy.)

Not sure if you want to group this with the related issue for supporting editing, #2179. IMO it would be valuable to have this notification make sense even without Pinafore supporting editing one's own toots. (Maybe there should be a third issue for viewing past versions of edited toots?)

graue:
Fixes #2280

toot-edit

I looked at "Inspect accessibility properties" in Firefox DevTools and the accessibility label looks right, similar to that for a boost notification.

Of course, these new strings aren't internationalized, as I'm not fluent in the relevant languages; let me know if there's anything you recommend I do about that.

### DIFF
--- a/src/intl/en-US.js
+++ b/src/intl/en-US.js
@@ -513,6 +513,7 @@ export default {
   rebloggedYou: 'boosted your toot',
   favoritedYou: 'favorited your toot',
   followedYou: 'followed you',
+  edited: 'edited their toot',
   signedUp: 'signed up',
   posted: 'posted',
   pollYouCreatedEnded: 'A poll you created has ended',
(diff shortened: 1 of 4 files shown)


nolanlawson:
Looks great, thanks. You don't need to edit the non-English files because we effectively only support English right now.

nolanlawson (approved):
Appreciate it! Thank you.

PR #2312, opened 2022-12-11
Prototype more elegant status icons

NickColley:
Just so you know... I tried having the retweet be an outline stroke like the star and it looks awful.


NickColley:
@nolanlawson what you think to this? If you're into it I'll get it ready to merge... (remove checked code etc) Builds off https://github.com/nolanlawson/pinafore/pull/2297

NickColley:
Closing this for now as want to try something else with the retoot too...

NickColley:
Yeah agreed, I like our current approach it feels much more intentional but will keep thinking on it

PR #2314, opened 2022-12-11
fix: show proper notification text for follow request

### LINKED ISSUE #1800: Follow request notifications unclear
Pending follow requests show up clearly in the Community tab as a number next to the entry labelled "Follow requests", but under Notifications, they show up for me as two entries for each request, one with the user's display name and one with user's fully-qualified address (e.g. `bob@example.net`), both with the user's profile picture and the five-pointed star icon to the left.  Clicking on either brings me to the user's profile instead of to the follow requests.  (Note, this is with a Pleroma instance; I am unsure whether this behaviour occurs with Mastodon.)

### LINKED ISSUE #2263: Follow requests are just the person's name
Hi, when someone requests to follow me, the notification I see in Pinafore is just the person's name. It doesn't say "...requested to follow you" or anything. It doesn't have a way to accept or reject, it only links to their profile.

In this screenshot, the notifications that are just a profile pic and name are follow requests:

pina

This is on a server running Mastodon 4.0.2. I saw an issue, #1800, that sounded related to this, but it appears to be somewhat different.

nolanlawson:
Fixes #1800

Shows the correct text "so-and-so requested to follow you" and an icon (hourglass).

This doesn't fully solve the problem, because clicking on it just takes you to the account. You still need to go to Community -> Follow requests to actually approve. But it's better than the status quo.


src/routes/_api/timelines.js
```diff
@@ -66,7 +66,7 @@ export async function getTimeline (instanceName, accessToken, timeline, maxId, s
   }
 
   if (timeline === 'notifications/mentions') {
-    params.exclude_types = ['follow', 'favourite', 'reblog', 'poll', 'admin.sign_up']
+    params.exclude_types = ['follow', 'favourite', 'reblog', 'poll', 'admin.sign_up', 'update', 'follow_request']
```
nolanlawson, line 69: Realized we forgot to add `'update'` here. There are so many notification types now (signup, update, follow request, report), this system is becoming untenable.

Issue #2179, opened 2022-11-12
Edit toot


nolanlawson:
I guess this is a new thing in Mastodon? Needs some way to determine whether or not the server supports it, though. E.g. I dunno if Pleroma supports this.

tbroyer:
Will apparently be in Pleroma 2.5.0: https://git.pleroma.social/pleroma/pleroma/-/merge_requests/3678

I suppose that one way to detect support is to look at the `version` field of `/api/v1/instance`, and look for a version >= 3.5.0 (or 4.0.0 for when the feature was enabled in Mastdon's web UI). Pleroma returns 2.7.2 currently so it wouldn't be detected as supporting editing. No idea what it'll return when they release 2.5.0 with editing feature so might have to revisit (will they still return `2.7.2 (compatible; Pleroma 2.5.0)`? In which case detection would have to look for the Pleroma version as well)

nolanlawson:
@tbroyer Thanks for the info. Usually what I do in these sorts of situations is look to other fediverse clients and see how they handle it. E.g. `max_toot_chars` used to be the way to communicate the number of max toot characters, but that changed: https://github.com/nolanlawson/pinafore/issues/2187

From a glance at https://mastodon.social/api/v1/instance though I don't see anything related to toot editing, so probably version-sniffing will be the only solution.

tbroyer:
The official Android application apparently doesn't do any detection/sniffing, whereas the iOS application doesn't support editing yet.

Tooot apparently uses version sniffing for everything: https://github.com/tooot-app/app/blob/bd750da9d5888cde07fb8f29ded9821695e1de07/src/helpers/features.json / https://github.com/tooot-app/app/blob/bd750da9d5888cde07fb8f29ded9821695e1de07/src/utils/slices/instancesSlice.ts#L326-L335

Same for iMast: https://github.com/cinderella-project/iMast/blob/931b265013aa0b11a683dba67f435bf3dd6bf4cd/Sources/iOS/App/CustomViews/Post/PostDetail/MastodonPostDetailReactionBarViewController.swift#L138-L142

Tusky will likely do version-sniffing: https://github.com/tuskyapp/Tusky/pull/2828#issuecomment-1322741526

Other apps I've looked at don't support editing yet.

jdaviescoates:
Yeah, I just came here to say that I miss having the new edit function when using Pinafore 

(and Tusky doesn't seem to have it yet either so missing it there too!)

[...]

PR #2325, opened 2022-12-17
fix: handle status edit events

nolanlawson:
This is step 1 in implementing editing (#2179). At the very least we can updatt the cache when a status is updated.


nolanlawson:
Well it's a start

Issue #513, opened 2018-08-28
Add aXe testing to automated Ci tests


nolanlawson:
We can use this testcafe plugin: https://github.com/helen-dikareva/axe-testcafe

sorin-davidoi:
The plugin uses an older version of aXe, but there are two pull requests opened if the latest version is needed.

nolanlawson:
Interestingly it seems aXe cannot test Pinafore. I've tried `axe-cli` as well, and it just hangs.
![screenshot from 2018-12-05 22-53-07](https://user-images.githubusercontent.com/283842/49566786-adf9eb80-f8e0-11e8-946f-2c5da680a2f3.png)

I thought maybe it was due to CSP, but I disabled it and it still fails.

On the bright side, Lighthouse does seem to be working so perhaps this should be closed in favor of https://github.com/nolanlawson/pinafore/issues/746

sorin-davidoi:
aXe runs for me in both Firefox Nightly and Chromium.

![image](https://user-images.githubusercontent.com/2109702/49657854-828a1480-fa38-11e8-8b9d-5b71d6e2e66e.png)

nolanlawson:
axe works for me now, although it's finding some false positives: https://github.com/dequelabs/axe-core/issues/1176#issuecomment-473704254

NickColley:
Want to get this going to ensure when we resolve contrast issues we dont accidentally regress so adding to milestone for WCAG compliance.

nolanlawson:
@NickColley Thanks a lot. AA sounds totally achievable for us. (My understanding is that AAA is more like a "stretch goal" so not necessarily something we need to target?)

NickColley:
Yeah some things you want to try and aim for AAA but sometimes it's actually impossible to meet all to AAA.

AA is what the main legislation for the EU and UK Public Sector Accessibility Regulations align with, and is becoming a sort of defacto baseline, and definitely achievable for us.

NickColley:
Just a note on false positives, it's pretty common for axe to throw up issues that are false positives so what we'd do in this case is disable the rule that is being thrown and document why. This may also be a good way to introduce the tests as well, in a sort of 'failing test' situation where we add axe and raise github issues for each disabled rule.

PR #2326, opened 2022-12-18
Add automated contrast checking

NickColley:
Seeing if we can get automated testing running in GitHub actions, we could add checks in other specs if this is viable.

As part of #513

### DIFF
--- a/.testcaferc.json
+++ b/.testcaferc.json
@@ -5,5 +5,10 @@
     "takeOnFails": false
   },
   "skipJsErrors": true,
-  "stopOnFirstFail": true
+  "stopOnFirstFail": true,
+  "clientScripts": [
+    {
+      "module": "axe-core/axe.min.js"
+    }
+  ]
 }
--- a/package.json
+++ b/package.json
@@ -119,7 +119,9 @@
     "worker-loader": "^3.0.8"
   },
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


.testcaferc.json
```diff
@@ -5,5 +5,10 @@
     "takeOnFails": false
   },
   "skipJsErrors": true,
-  "stopOnFirstFail": true
-}
+  "stopOnFirstFail": true,
+  "clientScripts": [
+    {
+      "module": "axe-core/axe.min.js"
```
nolanlawson, line 11:
```suggestion
      "module": "axe-core/axe.js"
```

Might help with debugging.

nolanlawson (approved):
:+1: when tests are passing (requires the other PRs I imagine). This is awesome.

There are also some lint issues but you should be able to fix that with `npx standard --fix`.

.testcaferc.json
NickColley: I wonder if the minified script runs quicker which might add up in the long run if we start using it more. But probably not.

nolanlawson: That's fair. If we end up needing to debug something, we can switch to `axe.js`.

NickColley:
Closing as unmaintained, much love Nolan

PR #2331, opened 2022-12-21
fix: lighten button colors on some themes

nolanlawson:
Tweaks the colors on the majesty/grayscale themes. The buttons are still >3.0 contrast ratio.


NickColley (approved):
Looking good! Nice one 👍

PR #2343, opened 2023-01-03
OAuth2: use correct `Content-Type` as specified in RFC

vitalyster:
* Token request should use `application/x-www-form-urlencoded`: https://www.rfc-editor.org/rfc/rfc6749#section-4.1.3

### DIFF
--- a/src/routes/_api/oauth.js
+++ b/src/routes/_api/oauth.js
@@ -5,7 +5,7 @@ const WEBSITE = 'https://pinafore.social'
 const SCOPES = 'read write follow push'
 const CLIENT_NAME = 'Pinafore'
 
-export function registerApplication(instanceName, redirectUri) {
+export function registerApplication (instanceName, redirectUri) {
   const url = `${basename(instanceName)}/api/v1/apps`
   return post(url, {
     client_name: CLIENT_NAME,
@@ -15,7 +15,7 @@ export function registerApplication(instanceName, redirectUri) {
   }, null, { timeout: WRITE_TIMEOUT })
 }
 
-export function generateAuthLink(instanceName, clientId, redirectUri) {
+export function generateAuthLink (instanceName, clientId, redirectUri) {
   const params = paramsString({
     client_id: clientId,
     redirect_uri: redirectUri,
@@ -25,7 +25,7 @@ export function generateAuthLink(instanceName, clientId, redirectUri) {
   return `${basename(instanceName)}/oauth/authorize?${params}`
 }
 
-export function getAccessTokenFromAuthCode(instanceName, clientId, clientSecret, code, redirectUri) {
+export function getAccessTokenFromAuthCode (instanceName, clientId, clientSecret, code, redirectUri) {
   const url = `${basename(instanceName)}/oauth/token`
   // Using URLSearchParams here guarantees a content type of application/x-www-form-urlencoded
   // See https://fetch.spec.whatwg.org/#bodyinit-unions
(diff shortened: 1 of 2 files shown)


src/routes/_utils/ajax.js
```diff
@@ -51,7 +51,7 @@ async function _fetch (url, fetchOptions, options) {
 async function _putOrPostOrPatch (method, url, body, headers, options) {
   const fetchOptions = makeFetchOptions(method, headers, options)
   if (body) {
-    if (body instanceof FormData) {
+    if (body instanceof FormData || body instanceof URLSearchParams) {
       fetchOptions.body = body
```
nolanlawson, line 55: It seems that, by default, since the `Content-Type` is not set here, the browser (Chromium at least) sets

- `multipart/form-data; boundary=----WebKitFormBoundaryCBpNkZBPd2t28viK` for `FormData`
-  `application/x-www-form-urlencoded; charset=utf-8` for `URLSearchParams`

This is really surprising behavior to me. To me it would make more sense to explicitly specify the two content-types in the two different cases.

src/routes/_api/oauth.js
```diff
@@ -27,11 +27,11 @@ export function generateAuthLink (instanceName, clientId, redirectUri) {
...
-    grant_type: 'authorization_code',
-    code
-  }, null, { timeout: WRITE_TIMEOUT })
+  const data = new FormData();
+  data.append('client_id', clientId);
+  data.append('client_secret', clientSecret);
+  data.append('redirect_uri', redirectUri);
+  data.append('grant_type', 'authorization_code');
+  data.append('code', code);
+  return post(url, new URLSearchParams(data), null, { timeout: WRITE_TIMEOUT })
```
nolanlawson, line 36: Based on my reading of [the `URLSearchParams` constructor](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams), it does not actually care if you use `FormData` – any iterable of key-value string pairs will work here. So I find it a bit confusing to be using `FormData` here.

nolanlawson:
Thanks for opening this PR! Could you explain what problem this solves? E.g. is there a fediverse server that is broken with Pinafore's current approach?

vitalyster:
I'm implementing Mastodon API on my own server and that is the issue I spend hours to find why Pinafore doesn't sign in. It's better to make it use standard behavior to help new developers in the future.

src/routes/_utils/ajax.js
vitalyster: That seems to be documented in https://fetch.spec.whatwg.org/#bodyinit-unions

nolanlawson: Fascinating, thanks!

src/routes/_api/oauth.js
```diff
@@ -25,13 +25,13 @@ export function generateAuthLink (instanceName, clientId, redirectUri) {
   return `${basename(instanceName)}/oauth/authorize?${params}`
 }
 
-export function getAccessTokenFromAuthCode (instanceName, clientId, clientSecret, code, redirectUri) {
+export function getAccessTokenFromAuthCode(instanceName, clientId, clientSecret, code, redirectUri) {
   const url = `${basename(instanceName)}/oauth/token`
-  return post(url, {
+  return post(url, new URLSearchParams({
```
nolanlawson, line 30:
```suggestion
  // Using URLSearchParams here guarantees a content type of application/x-www-form-urlencoded
  // See https://fetch.spec.whatwg.org/#bodyinit-unions
  return post(url, new URLSearchParams({
```

nolanlawson (approved):

